PR #8, opened 2016-11-12
Activityreport

johan-bjareholt:
Do not merge before
https://github.com/ActivityWatch/aw-core/pull/12
https://github.com/ActivityWatch/aw-server/pull/17


src/views/Activity.vue
```diff
@@ -0,0 +1,428 @@
...
+
+
+    /*
+
+        Date/time functions
+
+    */
+
+
+    setDay: function(datestr){
```
ErikBjare: We have [moment.js](http://momentjs.com/) as a dependency, it's a lot more reasonable to use.

johan-bjareholt: Started using momentjs now, saved me 27 lines :)

ErikBjare:
A bit is left to be desired in terms of code complexity and readability. But it'll do for now.

Merging.

PR #10, opened 2017-02-17
Started work on new design

ErikBjare:
Work in progress!

screenshot_2017-03-11_09-43-00

Inspiration:
 - https://github.com/syncthing/syncthing/issues/1381


ErikBjare:
We really should implement [the TODO specified here](https://github.com/ActivityWatch/aw-webui/blob/dev/new-design/src/App.vue#L24) (highlight the active link in menu, home/activity/buckets/log).

If we do that we might remove the \<h1\> tags from the subview.

johan-bjareholt:
Just tested, works really nice.

By the way, the repo seems to now be configured to use port 5666 by default, should we keep it like this or should we change it to 5600? The best way would be to make it run on 5666 when using npm run dev and 5600 when running npm run build, but i have no idea to do that.

src/resources.js
```diff
@@ -2,14 +2,14 @@ import Vue from 'vue';
...
 let hostname = "localhost";
-let port = "5600";
-//let port = "5666";
+//let port = "5600";
+let port = "5666";
 let origin = protocol + "://" + hostname + ":" + port;
- */
-let origin = "";
+
+//let origin = "";
```
johan-bjareholt: This is exactly what breaks so you cannot use both 127.0.0.1 and localhost. Before we simply used origin="" so it chose 127.0.0.1/localhost depending on what's in the URL so no CORS issues occur, but when hardcoding localhost you cannot use 127.0.0.1

ErikBjare:
@johan-bjareholt I implemented the port-picking logic depending on `npm run dev/build` here: https://github.com/ActivityWatch/aw-webui/pull/10/commits/8cc5b45a2850f046758adc67d9fc6783ba406de3

ErikBjare:
The latest commit https://github.com/ActivityWatch/aw-webui/pull/10/commits/d5278695a9a088b77173779be6032b79830cfb5d should fix an issue in the previous commit where host might not be `localhost` but instead `127.0.0.1` or a remote IP.

ErikBjare:
In the last commit I added display of when a bucket was last updated. This depends on the following PRs:

 - https://github.com/ActivityWatch/aw-server/pull/19 (merges into `dev/heartbeat`, not `master`)
 - https://github.com/ActivityWatch/aw-core/pull/19

If these PRs aren't merged, it will simply not show anything.

screenshot_2017-03-12_10-09-12

PR #12, opened 2017-03-13
Fixed export bucket (now doing a full export)

ErikBjare:
I've fixed this by setting `?limit=-1`, we should ensure that `limit=-1` really means **no limit** on the serverside before merging.


johan-bjareholt:
I have tested that -1 becomes no limit on aw-server before, but it was a while ago so it wouldn't hurt if you did it again.

johan-bjareholt:
Improved tests in aw-core for limit=-1 and limit=0

https://github.com/ActivityWatch/aw-core/pull/21

ErikBjare:
Great, merging.

PR #13, opened 2017-03-13
Timeline

ErikBjare:


src/util/time.js
```diff
@@ -0,0 +1,37 @@
+import moment from 'moment';
+
+module.exports = {
+    seconds_to_duration: function(seconds){
```
ErikBjare: Should probably be written using moment instead

PR #15, opened 2017-03-14
View dropdown and better activity error handling

johan-bjareholt:


johan-bjareholt:
I have used this for quite some time and it works fine. Maybe not the best solution with the dropdown menu but better than nothing. @ErikBjare Should i merge or should we wait until we find out a better solution for it?

ErikBjare:
@johan-bjareholt You may go ahead and merge.

PR #24, opened 2017-05-29
Minor cleanup to visualization code

ErikBjare:


ErikBjare:
Some more work to be done here.

johan-bjareholt:
Why not merge it for now? The changed homescreen should be a nice improvement.

ErikBjare:
Pretty sure the timeline is broken, I cherrypicked the home view text just an hour ago so it's now in master.

ErikBjare:
This now works as it should. Just awaiting review from @johan-bjareholt.

src/visualizations/timeline.js
```diff
@@ -17,24 +17,23 @@ var time = require("../util/time.js");
  * titleinfo-container (visible titleinfo)
  * */
 
+// Helper functions used when hover state changes.
+function set_color(elem_id, color) {
```
johan-bjareholt: Is it possible for the on-click events to call these? If so, awesome!
The reason why I inserted these functions with a <script> element was because i thought they would be out of scope otherwise.

ErikBjare: Nope, will work fine :+1:

PR #29, opened 2017-07-17
Visualization status and refactor

johan-bjareholt:
Visualization code is still kind of a mess, but are at least slightly cleaner now and show "Loading..." text while doing REST requests to be more "reactive"

- Added "Loading..." text while doing REST request
- Replaced timeline titleinfo svg with a html table
- Visualization components now work with `create`, `update` and `set_status` methods instead of a single `render`
- Misc refactoring of visualizations

### DIFF
--- a/src/views/Activity.vue
+++ b/src/views/Activity.vue
@@ -22,7 +22,7 @@ h4 Summary
 
 h5 Total time: {{ time.seconds_to_duration(duration) }}
 
-div#appsummary
+div#appsummary_container
 
 hr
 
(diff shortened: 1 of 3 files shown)


src/views/Activity.vue
```diff
@@ -22,13 +22,13 @@ h4 Summary
 
 h5 Total time: {{ time.seconds_to_duration(duration) }}
 
-div#appsummary
+div#appsummary_container
```
ErikBjare: Please use underscores for ID's. Same goes for apptimeline_container.

ErikBjare:
Definitely an improvement, just make all CSS IDs and classes use dashes instead of underscores and I'm happy.

Will make me have to fix merge conflicts for https://github.com/ActivityWatch/aw-webui/pull/24 but it's my own fault for postponing it ¯\_(ツ)_/¯

PR #33, opened 2017-08-01
Made some improvements to styling needed after upgrading packages

ErikBjare:


src/views/Activity.vue
```diff
@@ -6,14 +6,12 @@ div
...
 
-  h4(style="color: red;")
-    | This is currently work in progress and is known to have issues (such as
-    a(href='https://github.com/ActivityWatch/aw-webui/issues/22')  needing a date selector
-    | ,
-    a(href='https://github.com/ActivityWatch/activitywatch/issues/72')  timezone issues
-    | , etc.) - see
-    a(href='https://github.com/ActivityWatch/aw-webui/issues')  all issues
-    | .
+  //div.alert.alert-info(role="alert")
```
johan-bjareholt: Why is this hidden?
I commented it out myself and the list seems to be badly positioned, but looks much better otherwise.
Also, I don't think you should remove the 'timezone issues' and 'all issues' links

src/views/Activity.vue
```diff
@@ -42,14 +40,22 @@ div
...
+    | Events queried: {{ eventcount }}
 
 </template>
 
 <style lang="scss">
 
+#apptimeline-container {
+    white-space: nowrap;
+    font-family: arial;
+    font-size: 10pt;
```
johan-bjareholt: This is a bit too small in my opinion. I personally like small fonts, but we should consider others with slightly worse than average vision and have 11 or 12pt as a minimum everywhere.

src/views/Activity.vue
ErikBjare: I wanted to keep the code in case we want to use it in other places.

Why not remove a link to the timezone issue? I fixed the basic issue and we settled that the remaining issues involving timezones won't be solved for quite a while, same for the datepicker.

The 'all issues' link isn't of much use since there aren't many issues that users need to concern themselves with.

src/views/Activity.vue
ErikBjare: That's a fair point, will fix.

PR #39, opened 2017-08-13
A lot of different stuff

ErikBjare:
I got lazy and didn't write clear commit messages.

Stuff that's changed:

 - Added moment-range as a dependency
 - Implemented `filterPeriodIntersect` (same behavior as `filter_period_intersect` in aw-core) using moment-range.
 - The Activity view now also has a "most common window titles" that uses `filterPeriodIntersect`.
 - I moved away from using the view API in the "most common apps" chart, it now uses `filterPeriodIntersect` instead.
 - Colors are now predictable (thanks to hashcode and d3.js)
 - A bunch of cleanup.

Things left:
 - [ ] I will probably rewrite the timeline thing as well to not depend on the view API.

### DIFF
--- a/src/util/color.js
+++ b/src/util/color.js
@@ -1,39 +1,41 @@
-const Color = require("color");
+const d3 = require("d3");
 
-/*
-    Color generation
-    ================
-    Each app has its own color in app_colors
-    The first 4 apps get the 4 default colors
-    The preceeding apps get a mix in the order of 1+0,2+0,2+1,3+0,3+1,3+2...
-*/
-var colors = [Color('#90CAF9'), Color('#FFE082'), Color('#EF9A9A'), Color('#A5D6A7')];
-let app_colors = {};
+// See here for examples:
+//   https://bl.ocks.org/pstuffa/3393ff2711a53975040077b7453781a9
 
-function calculateNextColor() {
-  var color1 = colors[(colors.length >> 1)-1];
-  var color2 = colors[colors.length%(colors.length >> 1)];
+let scale = d3.scaleOrdinal(d3.schemeCategory20c);
 
-  // Now mix the two
-  var color = Color({
-    r: (color1.rgb().r+color2.rgb().r)/2,
-    g: (color1.rgb().g+color2.rgb().g)/2,
-    b: (color1.rgb().b+color2.rgb().b)/2
-  });
+// Needed to prewarm the color table
+scale.domain([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]);
 
-  colors[Object.keys(app_colors).length] = color;
+let customColors = {
+    "afk": "#DDD",
+    "not-afk": "#2D6",
+
+    "chrome": "#4885ed",   // Google Blue
+    "chromium": "#4885ed", // Google Blue
+    "spotify": "#1ED760",  // Spotify Green
+    "alacritty": "#EA5"
 }
 
-function getAppColor(appname) {
-  if(!(appname in app_colors)) {
-    if (colors.length == Object.keys(app_colors).length){
-      calculateNextColor();
+function hashcode(str){
+    var hash = 0;
+    if (str.length == 0) return hash;
+    for (var i = 0; i < str.length; i++) {
+        var character = str.charCodeAt(i);
+        hash = ((hash<<5)-hash)+character;
+        hash = hash & hash; // Convert to 32bit integer
     }
-    app_colors[appname] = colors[Object.keys(app_colors).length]
-  }
-  return app_colors[appname].rgbString();
+    return hash;
+}
+
+function getColorFromString(appname) {
+    appname = appname || "";
+    appname = appname.toLowerCase();
+    return customColors[appname] || scale(Math.abs(hashcode(appname) % 20));
 }
 
 module.exports = {
-    getAppColor: getAppColor
+    getAppColor: getColorFromString,
+    getColorFromString: getColorFromString
 };
--- a/src/views/Activity.vue
+++ b/src/views/Activity.vue
@@ -52,7 +52,7 @@ div
   hr
 
   p
-    //| Showing activity from {{ datestart }} until 24 hours later
+    //| Showing activity from {{ dateStart }} until 24 hours later
     //br
     | Events queried: {{ eventcount }}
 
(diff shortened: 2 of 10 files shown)


src/util/color.js
```diff
@@ -1,39 +1,41 @@
...
-
-  // Now mix the two
-  var color = Color({
-    r: (color1.rgb().r+color2.rgb().r)/2,
-    g: (color1.rgb().g+color2.rgb().g)/2,
-    b: (color1.rgb().b+color2.rgb().b)/2
-  });
-
-  colors[Object.keys(app_colors).length] = color;
+const d3 = require("d3");
```
johan-bjareholt: The code is a lot nicer and app-specific colors is a great feature, but the colors look awful on my computer. The only apps i use are firefox and urxvt and they are a light brown-yellow and a bright orange color respectively.

ErikBjare: Yeah, the colors are hardly final. I just picked a color set included with d3.js and didn't feel like bikeshedding over colors at the time. Feel free to find a set you like better.

src/views/Today.vue
```diff
@@ -61,13 +61,15 @@ export default {
   },
 
   getEvents: function(bucket_id, start, end) {
-    let limit = 1000;
-    return $Event.get({"id": bucket_id, "start": start, "end": end, "limit": 1000}).then((response) => {
+    let limit = 10000;
```
ErikBjare: The limit was just there for testing purposes, I will remove it.

src/views/Activity.vue
```diff
@@ -91,158 +98,129 @@ export default {
...
+    this.query();
   },
 
   methods: {
-    previousDay: function() {
-        return moment(this.datestart).subtract(1, 'days').format("YYYY-MM-DD");
-    },
-    nextDay: function() {
-        return moment(this.datestart).add(1, 'days').format("YYYY-MM-DD");
+    previousDay: function() { return moment(this.dateStart).subtract(1, 'days').format("YYYY-MM-DD") },
```
ErikBjare: I just got annoyed by the long file and felt that it could be onelined since you should never really need to read it. Might be better to turn it into a computed property.

ErikBjare:
@johan-bjareholt It appears that lodash wasn't added as a dependency in `package.json`. Likely to have been the issue. Fixed in latest commit.

Regarding the client-side views: I think the server-side views is way too messy and this was the only way for me to do any work at all. I just couldn't work with the views API, it's too much work if you want to change something and the whole querying thing is a nightmare. Doing the transforms with lodash was easy and the clientside code is a lot easier to maintain than an API. I hardly consider it a waste of time. The new code is unarguably a lot clearer.

The long term arguments for having it all server side are still valid but with the current feature set there is no actual need. This is a lot simpler and not relying on it for now would make it easier to rewrite the view API (once we have an actual need for one). The plan I have is to deprecating the view API and move the transforms into aw-analysis and then from there start investigating a new design for the view API **when it's needed**. Before then I can't justify the complexity nor do I believe we can accurately anticipate our needs then.

>Also, you already know what i think that rewriting all transforms on the clientside with javascript is the wrong way to go

In the long run, sure. But I strongly disagree about that being the case in the short run. This way it's a lot easier to develop new visualizations.

>...and a waste of time.

The reason I ended up doing it was that I threw together the window view and realized it did almost exactly the same thing as the app view. So I saw an opportunity to rid ourselves of our dependency on the messy view API and extracted it in a couple of minutes. The whole thing (colors and other cleanup excluded) took less than an hour and brought a lot clearer code. I hardly consider it a waste of time.

>We already have unit tested transforms in aw-core and we both agree that doing it server-side is better in the long run.

The unit testing is still useful for when we move the transforms to aw-analysis. I fail to see why doing it server side is better **currently**.

johan-bjareholt:
> Doing the transforms with lodash was easy and the clientside code is a lot easier to maintain than an API. I hardly consider it a waste of time. The new code is unarguably a lot clearer.

That's your opinion. I agree that the views are not perfect and need a rewrite, but in it's core it's just a chain of transforms so its nothing complicated. Personally i don't understand much of all this javascript code because

1. It's javascript
2. It has no clear input->output which the python transforms had
3. I'm not a fan of writing complex code in javascript due to how error prone it is and lacks proper error reporting and exceptions

Personally I don't understand much of the code you have written. The reason why it's a waste of time is because before I understood it and now you understand it so the only thing we have done by this is switching who has experience with the code base.

ErikBjare:
After our discussion afk yesterday I've decided to continue on this path for now. I'll revisit the server side views sometime, likely to implement them as fixed transforms through the API rather than a kind of query language.

## Note: I wrote this before talking to you yesterday, feel free to not read.

>That's your opinion.

Not really, by every objective measure I can come up with (cyclomatic complexity, linecount, depth of the structures passed around) this is less complex. You might not think it's clearer, but I'd say that's just because you haven't gotten used to underscore/lodash.

Concerning your points:

 1. It has no clear input->output which the python transforms had
    - The transforms in Python are **nested dicts**, there's nothing clear about that. The lodash transforms in comparison are pretty standard and just work with lists of events (mostly).
 2. I'm not a fan of writing complex code in javascript due to how error prone it is and lacks proper error reporting and exceptions.
    - lodash works fine, I don't think your general "JavaScript is a weakly typed mess" applies here. Not sure what error reporting and exceptions has to do with this, but `console.error` works good enough for me.
 3. I don't want to repeat myself, and if we do views/queries properly it can be very powerful
    - It's not much actually being repeated. The bulk of the code in the Python code comes from the views/queries code, the transform wrappers, and the chunking functions. The views/queries wouldn't be repeated and the chunking functions could be written a lot shorter (and with lower cyclomatic complexity) using lodash.
    - I'm not sure what you are referring to when you say that views/queries could be very powerful. The only real value they add is caching? Any reasonably simple (and not overly complex) implementation would come at the cost of flexibility.

>Because we have it working

It's working, but it's not workable. Especially not so when developing new visualizations.

>there's no reason to switch to client-side just because you don't understand the server-side implementation. The reason why it's a waste of time is because before I understood the transforms implementation server-side and now you understand the transforms implementation client-side so the only thing we have done by this is switching who has experience with the code base.

I understand the transforms, that's not the issue, the issue is that the whole views/querying thing is overly complex and passes around huge nested dicts that a long way down the nesting resolves to a database query. Adding the complexity of an API, a database table, and several layers of abstraction around the actual transforms (I count ~3) isn't worth a little caching on the server side right now.

>Feel free to rewrite the whole server side queries and views if you want

I considered it and realized it would be a **lot** of work to get right and there is no current need, so I didn't.

>I don't want to mess around with complex javascript and lose future possibilities.

The JavaScript isn't complex in comparison. There are no possibilities lost.

johan-bjareholt:
> The transforms in Python are nested dicts, there's nothing clear about that. The lodash transforms in comparison are pretty standard and just work with lists of events (mostly).

The transforms are not nested dicts, the queries are.

> chunking functions could be written a lot shorter (and with lower cyclomatic complexity) using lodash.

The lodash chunker and the python chunker doesn't have the same functionality. It would be trivial to rewrite the chunker with the same functionality as yours in python with the same cyclomatic complexity.

> I'm not sure what you are referring to when you say that views/queries could be very powerful. The only real value they add is caching? Any reasonably simple (and not overly complex) implementation would come at the cost of flexibility.

Caching is incredibly important to me. Why would a year of data be interesting if you cannot see a proper overview of it? I don't have time to go back day-by-day and look at my activity a month ago, so if we have no summary I could just as well just scrap all data older than a few weeks.

> It's working, but it's not workable. Especially not so when developing new visualizations.

Your opinion, I believe the reverse.

> I understand the transforms, that's not the issue, the issue is that the whole views/querying thing is overly complex and passes around huge nested dicts that a long way down the nesting resolves to a database query. 

Here's how it possibly could look hardcoded in aw-server

```
    @check_bucket_exists
    def get_windowactivity(self, host: str, limit: int = -1,
                   start: datetime = None, end: datetime = None) -> List[Event]:
        """Get windowactivity events from a specific host"""
        windowevents = self.db[bucket_id + "_windowactivity"].get(limit, start, end)
        afkevents = self.db[host + "_afkstatus"].get(limit, start, end)
        events = transforms.filter_keyvals(windowevents, "afkstatsu", "not-afk", afkevents)
        return events
```
Way cleaner in my opinion.
Add a sort by duration on the front-end and then you have the same funcitonality, but unit tested.

ErikBjare:
>The transforms are not nested dicts, the queries are.

Woops, wrong word. I meant queries, I think I said transforms because that's what the queries are used to call (rewrote that sentence once and it slipped through).

>It would be trivial to rewrite the chunker with the same functionality as yours in python with the same cyclomatic complexity.

That would be nice, it's not that hard to rewrite the JS chunker to do what is done in the Python chunker though. Lodash is pretty amazing, the same things could be done in Python and would probably be rewritten that way.

>Your opinion, I believe the reverse.

All I'm saying: I think if you learned lodash well you too would find it less complex.

>Caching is incredibly important to me. Why would a year of data be interesting if you cannot see a proper overview of it? I don't have time to go back day-by-day and look at my activity a month ago, so if we have no summary I could just as well just scrap all data older than a few weeks.

It could still be done, you just had to wait a maximum of 365 times as long as for a single day every time you wanted to. That's ~6 minutes if we assume 1s per day with some pretty conservative assumptions. A less conservative estimate would be less than 2 min after optimizations. Waiting 2-6 min for something that you'll probably not do very often wouldn't be a biggie imo. A year-long query would take about as long in aw-server (assuming Python has roughly the same performance as JS in these tasks, which I'd expect). Sure, the server would cache it, but it's not requested that often so the benefit doesn't add up to much.

>Here's how it possibly could look hardcoded in aw-server:
>
>\<great example\>

@johan-bjareholt That's a really nice example, I like that :+1:. I generally believe most of the complexity would be easy to prune once we will get rid of the current querying, that was the major blocker. I have more hope for the server-side transforms if we abandon it.

**We're going to do what we've come up with in the discussion (predefined queries)**

------------------

**Here is something else, but slightly related,** that has been in the back of my mind:

One slightly related long term concern I have is this: A simple server implementation is to be preferred over a simple Web UI. The server could benefit from being reimplemented in another language (such as Rust, see: https://github.com/ActivityWatch/activitywatch/issues/31). 

Such an implementation would be extremely efficient even with frequent heartbeats, would open up for use on {CPU, memory, battery} constrained devices like phones and minimizes battery impact for things like laptops, IoT-stuff, drones, etc. 

`aw-server` is currently pretty lean (imo, in spec at least) and a reimplementation could be done pretty quickly in Rust I think (~2 months if worked on fully, more realistically ~6 due to sporadic development). 

But I've realized this isn't overly important to the current debate. In the event that we would like to rewrite the core parts in Rust and needed the we could still call transforms in Python since that's an irregular thing which makes it less important to optimize for performance (since it runs less often). **My current plan in case we want to do this is to make the `info` endpoint inform if it's transforms-capable or not, which could make light servers easy to implement.** 

All this might seem a bit far fetched, but I really want to keep simplicity when specifying what the essential functionality of core parts should be.

I believe it is important to be able to write clean specs for the things that matter the most (aw-server) while things like visualization don't need to be as strictly spec'd since it is rarely called from other code. Transforms are somewhere in between, but I think a query language is an example of something that's too much specification work for the benefit you net.

#### \</rant\>

ErikBjare:
I'm merging this for now. The aw-server/transforms work will come later.

johan-bjareholt:
@ErikBjare By the way, if you are so in love with lodash you can use this instead on the server https://github.com/dgilland/pydash

PR #42, opened 2017-08-17
Node 8 support

ErikBjare:
Found this: https://github.com/sass/node-sass/issues/2012

It was a nice find, lets just hope Travis passes :crossed_fingers:


ErikBjare:
The Node 8 version seems to finally run without warnings now, nice.

PR #48, opened 2017-10-07
WIP: transform-next

johan-bjareholt:
Replaced all javascript transformation with query2 requests and removed the last View stuff

- [x] Fix timeline being inconsistent when "Show AFK time" is toggled
- [x] Fix timeline being broken when "Show AFK time" isn't toggled
- [x] Non-hardcoded web bucket


ErikBjare:
We got this new issue today, so I'm wondering if you've tested that data is fetched using the right timezone? (you can check this by seeing if data is properly queried around midnight)

ErikBjare:
I think your code is likely to suffer from the same issue, I think I fixed it in this PR so you shouldn't have to worry about it: https://github.com/ActivityWatch/aw-webui/pull/49

ErikBjare:
By the way, how is "Show AFK time" broken/inconsistent? 

It had issues before you started working on query2. I'm fine with just adding a notification that its partially broken for now if it's something you don't feel like putting a lot of work into.

johan-bjareholt:
The start- and endtime should be the same as previous queries so it shouldn't differ, so i should probably rebase to your commit later.

The "Show AFK time" is inconsistent due to the timeline used the "total_duration" field in chunks, and since we now use the query2 merge function instead which doesn't have the total_duration field i did a second "total_duration" query. Generally that requests returns before the timeline query, but not always and in that case it can completely break so you have to refresh to fix it. Could be a trivial fix, but we'll see.

ErikBjare:
Can't you do the `total_duration` calculation client-side? Should be easy enough... Seems a bit overkill to make a query for it.

johan-bjareholt:
Lol, github doesn't like rebases for sure, commit history looks insane

src/views/Activity.vue
```diff
@@ -345,63 +346,58 @@ export default {
...
-ENDTIME="'+endtime+'" \n\
-not_afk=query_bucket("'+afkbucket+'") \n\
-events=query_bucket("'+windowbucket+'") \n\
-not_afk=filter_keyval(not_afk, "status", "not-afk", FALSE) \n\
-events=filter_period_intersect(events, not_afk) \n\
-events=sort_by_timestamp(events) \n\
-RETURN=events';
+    windowTimelineQuery: function(windowbucket, afkbucket){
+      return { "query": [
+        'not_afk = query_bucket("'+afkbucket+'");',
```
ErikBjare: All variables used in the query need to be sanitized, we should at the very least add a TODO/FIXME comment.

johan-bjareholt: Done

ErikBjare:
Merging this!

PR #49, opened 2017-10-16
Fixed preservation of timezone in handling of datetimes

ErikBjare:
I think this fixes the following issue: https://github.com/ActivityWatch/activitywatch/issues/117

The reason is that `new Date().toISOString()` returns a datetime normalized to UTC with the 'Z' suffix. When this is later used to construct `moment` objects the timezone information is lost. `moment().format()` preserves the timezone information in the string using the `+HH:MM` suffix.

### DIFF
--- a/src/views/Activity.vue
+++ b/src/views/Activity.vue
@@ -140,11 +140,11 @@ export default {
   computed: {
     readableDuration: function() { return time.seconds_to_duration(this.duration) },
     host: function() { return this.$route.params.host },
-    date: function() { return this.$route.params.date || new Date().toISOString() },
+    date: function() { return this.$route.params.date || moment().format() },
     dateStart: function() { return moment(this.date).startOf('day').format() },
     dateShort: function() { return moment(this.date).format("YYYY-MM-DD") },
-    windowBucketId: function() { return "aw-watcher-window" + (PRODUCTION ? "_" : "-testing_") + this.host },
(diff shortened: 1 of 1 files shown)


src/views/Activity.vue
```diff
@@ -140,11 +140,11 @@ export default {
...
     readableDuration: function() { return time.seconds_to_duration(this.duration) },
     host: function() { return this.$route.params.host },
-    date: function() { return this.$route.params.date || new Date().toISOString() },
+    date: function() { return this.$route.params.date || moment().format() },
     dateStart: function() { return moment(this.date).startOf('day').format() },
     dateShort: function() { return moment(this.date).format("YYYY-MM-DD") },
-    windowBucketId: function() { return "aw-watcher-window" + (PRODUCTION ? "_" : "-testing_") + this.host },
-    afkBucketId:    function() { return "aw-watcher-afk"    + (PRODUCTION ? "_" : "-testing_") + this.host }
+    windowBucketId: function() { return "aw-watcher-window_" + this.host },
+    afkBucketId:    function() { return "aw-watcher-afk_"    + this.host }
```
ErikBjare: These changes also came along in the `git add`. 

They ensure compatibility with the following change: https://github.com/ActivityWatch/aw-client/commit/516bf1dadde2ce86c8bccc7785145cf216631ab1

ErikBjare:
I'm merging this without review, should be an uncontroversial change.

PR #52, opened 2017-11-30
Bumped version of node-sass to 4.7.2

ahnlabb:
- node-sass 4.5.3 does not support node 9


ErikBjare:
Looking good, nice with Node 9 support. Merging.

PR #58, opened 2018-02-11
Activity button instead of dropdown if one host

ianobermiller:
If there is only one host, don’t bother making Activity be a dropdown;
just link directly to the first host.

image


ErikBjare:
Nice work! 🎉 

Merging.

PR #60, opened 2018-02-24
New Query2 timeperiod api and monthly activity diagram

johan-bjareholt:


src/visualizations/periodusage.js
```diff
@@ -0,0 +1,116 @@
...
+          var usage_time = day_events[di].duration;
+          return usage_time;
+        }
+      }
+    }
+    return 0;
+  }
+
+  var longest_usage = 0.00000000001; // Not zero to avoid division by zero
+  for (var i in usage_arr){
```
ErikBjare: Or just do something like: `max(map(get_usage_time, day_events))`

src/visualizations/periodusage.js
```diff
@@ -0,0 +1,116 @@
...
+  if (usage_arr.length <= 0){
+    set_status(svg_elem, "No data");
+    return svg;
+  }
+  svg_elem.innerHTML = "";
+  var svg = d3.select(svg_elem);
+
+  function get_usage_time(day_events){
+    if (day_events.length >= 1){
+      for (var di in day_events){
```
ErikBjare: Or do: 
```
day_event = _.head(_.filter(day_events, (e) => day_event.data.status == "not-afk"));
return day_event != undefined ? day_event.duration : 0;
```

src/queries.js
```diff
@@ -1,68 +1,89 @@
...
-  return { "query": [
-    'not_afk = query_bucket("'+afkbucket+'");',
-    'events  = query_bucket("'+windowbucket+'");',
-    'not_afk = filter_keyvals(not_afk, "status", "not-afk");',
-    'events  = filter_period_intersect(events, not_afk);',
-    'events  = sort_by_timestamp(events);',
-    'RETURN  = events;',
+function windowTimelineQuery(starttime, endtime, windowbucket, afkbucket){
+  return {
+    "timeperiods": [starttime+"/"+endtime],
```
ErikBjare: Any particular reason why you opted for joining them into a string? Why not just a list-pair?

src/resources.js
```diff
@@ -19,7 +19,7 @@ let $Info       = Vue.resource(origin + '/api/0/info');
 let $Bucket     = Vue.resource(origin + '/api/0/buckets/{id}?force=1');
 let $Event      = Vue.resource(origin + '/api/0/buckets/{id}/events');
 let $EventCount = Vue.resource(origin + '/api/0/buckets/{id}/events/count');
-let $Query      = Vue.resource(origin + '/api/0/query/?name={name}&start={start}&end={end}&cache={cache}');
```
ErikBjare: Would it be much work to keep the old way of doing things in addition to the new one? (being able to specify period to query by URL parameters if you have only one period to query)

src/visualizations/periodusage.js
```diff
@@ -0,0 +1,116 @@
...
+  for (var i in usage_arr){
+    var usage_time = get_usage_time(usage_arr[i]);
+    var height = 100*(usage_time / longest_usage);
+    var date = "";
+    if (usage_arr[i].length > 0){
+      /* slice off so it's only the day */
+      date = usage_arr[i][0].timestamp.slice(0,10);
+    }
+    var color = diagramcolor;
+    if (i == center_elem){
```
ErikBjare: `var color = (i == center_elem) ? center_diagramcolor : diagramcolor;`

src/queries.js
johan-bjareholt: I did due to there apparently being a iso8601 timeperiod spec.

src/resources.js
johan-bjareholt: Not really, I just think it would be unnecessary to have two ways to do one thing without any significant benefit.

src/visualizations/periodusage.js
johan-bjareholt: Fixed

src/visualizations/periodusage.js
johan-bjareholt: Fixed

src/visualizations/periodusage.js
johan-bjareholt: Fixed

src/queries.js
ErikBjare: Oh wow, nice. I had no idea, I obviously support ISO8601 so no problem then. :smiley:

src/resources.js
ErikBjare: There is some benefit when it comes to the OpenAPI/Swagger spec, and therefore the API Browser. If it's just a couple lines I'd appreciate if you readded support for it in aw-server :+1:

src/queries.js
johan-bjareholt: That timeperiod feature in the spec is not implemented in the python package however so I am simply splitting with a "/" in the aw-server code. Not happy about that, but it does what it's supposed to.

src/resources.js
johan-bjareholt: > There is some benefit when it comes to the OpenAPI/Swagger spec, and therefore the API Browser. 

Well, the "timeperiods" field is still marshalled so we still don't break the OpenAPI/Swagger spec in any way as far as i understand. It just doesn't make sense to me to have two options in an API which depend or whether you want to provide one or more of a data type. An API should be rather straight forward just by looking at the definition, you shouldn't have to explain it with too much text. In this case we would have to document both options, document which one is prefered over the other in different situations, document which one is considered first priority if both are specified etc. which just screams bad design to me.

ErikBjare: Fair enough.

PR #62, opened 2018-03-28
Initial aw-client-js support

johan-bjareholt:
Needs aw-client-js cloned to the root folder to work.

Needs the dev/axios branch of aw-client-js to work https://github.com/ActivityWatch/aw-client-js/pull/2

PR #64, opened 2018-04-01
Added query explorer

ErikBjare:
Stuff left to do:

 - [ ] Better default query (automatically pick a suitable bucket?)
 - [x] Add returned event count
 - [x] Add basic visualization/result browser
 - [x] Merge #62 and use aw-client-js instead


ErikBjare:
Merging this, it'll improve over time but good enough for now.

PR #71, opened 2018-05-26
Updated NPM packages and other misc fixes

johan-bjareholt:
- Webpack 4
- Updated webpack configuration
- New font awesome version
- Fixed some vulnerability warning
- Removed shield.io for twitter/github due to it not working when
  offline
- Added two icons to Buckets view
- Removed node5 from travis (seems to be broken on webpack 4)


johan-bjareholt:
`npm run dev` works but `npm run build` is currently broken, webpack production conf is bad

PR #76, opened 2018-06-16
Added editor activity

johan-bjareholt:
Todo:
- [ ] Some minor styling fixes
- [ ] Test with multiple editor buckets
- [ ] Refactoring

### DIFF
--- a/src/queries.js
+++ b/src/queries.js
@@ -48,6 +48,20 @@ function browserSummaryQuery(browserbucket, windowbucket, afkbucket, count, filt
   ]);
 }
 
+function editorActivityQuery (editorbucket, limit){
+  return [
+    'editorbucket = "' + editorbucket + '";',
+    'events = flood(query_bucket(editorbucket));',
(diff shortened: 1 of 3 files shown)


johan-bjareholt:
Have now also tested with aw-watcher-vscode with some modifications

https://github.com/Otto-AA/aw-watcher-vscode/pull/2

Otto-AA:
Can't test because the app can't connect to the aw-server.
It fails at this part:
```javascript
awclient.info().then(
   (response) => {
     if (response.status > 304) {
       console.error("Status code from return call was >304");
     } else {
       this.connected = true;
       this.info = response.data;
     }
```
because it somehow makes the request with the method `OPTIONS` instead of `GET`. (Tested resending the request with `GET` and got the correct response). From the aw-client-js it seems to me that it _should_ make a get request.
A screenshot of the request (180kb is sent request, 0kb is response)
image

Do you have any ideas how to fix this?

PR #77, opened 2018-06-17
Made activity view categorized by tabs and added top URL summary

johan-bjareholt:
Made activity view tabbed.

We need some way to hide stuff such as editors which will not be used by everyone since the page will otherwise be very long, but not sure if this is the perfect solution.

screenshot-2018-6-17 activitywatch

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -11,9 +11,8 @@ div#wrapper
           | Connected
           icon(name="check-circle")
         span.bad(v-show="!connected")
-          span.text
-            | Not connected
-            icon(name="times-circle")
+          | Not connected
+          icon(name="times-circle")
 
   div.container.aw-container
     // TODO: Refactor into Mainmenu component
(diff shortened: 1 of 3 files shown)


ErikBjare:
Looks pretty good, should probably make the tabs stand out more though because I didn't see them at first.

johan-bjareholt:
screenshot-2018-6-18 activitywatch
There, they should be more clear now!
Required some custom CSS sadly.

Otto-AA:
Looks good :+1: 

If it's not too much coding effort, I think it would be awesome if one could choose which data is displayed in the summary. For instance, that it displays "Top coding Projects" instead of "Top window Titles".

johan-bjareholt:
Nice idea!

However that would be quite a bit of work due to the website currently being stateless, this would require us to save these settings somehow (probably either in the backend or in a cookie).
Personalization is probably what would make the user interface "great" instead of "good" though, so we will have to implement that eventually.

PR #84, opened 2018-06-20
Switch to bootstrap tabs menu

Otto-AA:
Thought I'd try to make the tab menu use bootstrap instead of custom css. It works, but I am not 100% happy with the aesthetics now. So if you think it looks not good enough to merge, I totally understand it ^_^

Here is a screenshot
image

### DIFF
--- a/src/views/Activity.vue
+++ b/src/views/Activity.vue
@@ -30,127 +30,120 @@ div
 
   aw-periodusage(:periodusage_arr="daily_activity", :host="host")
 
-  ul.nav.nav-tabs
-    li.nav-item.aw-nav-item
-      a.nav-link.aw-nav-link(v-on:click="view = 'summary'" v-bind:class="{ active: view=='summary' }")
-        h5 Summary
-    li.nav-item.aw-nav-item
-      a.nav-link.aw-nav-link(v-on:click="view = 'window'" v-bind:class="{ active: view=='window' }")
-        h5 Window
-    li.nav-item.aw-nav-item
-      a.nav-link.aw-nav-link(v-on:click="view = 'browser'" v-bind:class="{ active: view=='browser' }")
-        h5.active-h5 Browser
-    li.nav-item.aw-nav-item
-      a.nav-link.aw-nav-link(v-on:click="view = 'editor'" v-bind:class="{ active: view=='editor' }")
-        h5 Editor
-  br
+  // Nav tabs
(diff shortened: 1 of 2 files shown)


src/views/Activity.vue
```diff
@@ -30,127 +30,120 @@ div
...
-
-  div.row(v-show="view == 'summary'")
-    div.col-md-4
-      h5 Top Applications
-      aw-summary(:fields="top_apps", :namefunc="top_apps_namefunc", :colorfunc="top_apps_colorfunc")
-      b-button(size="sm", variant="outline-secondary", :disabled="top_apps.length < top_apps_count", v-on:click="top_apps_count += 5; queryWindows()")
+  // Nav tabs
+  ul.nav.nav-tabs.nav-pills.lead.mb-3.mt-4
+    li.nav-item
+      a.nav-link.active(data-toggle='tab' href='#summary') Summary
```
johan-bjareholt: data-toggle, nice!

src/views/Activity.vue
```diff
@@ -30,127 +30,120 @@ div
...
-        h5 Editor
-  br
-
-  div.row(v-show="view == 'summary'")
-    div.col-md-4
-      h5 Top Applications
-      aw-summary(:fields="top_apps", :namefunc="top_apps_namefunc", :colorfunc="top_apps_colorfunc")
-      b-button(size="sm", variant="outline-secondary", :disabled="top_apps.length < top_apps_count", v-on:click="top_apps_count += 5; queryWindows()")
+  // Nav tabs
+  ul.nav.nav-tabs.nav-pills.lead.mb-3.mt-4
```
johan-bjareholt: The reason why I went with custom CSS was because the normal tabs were not very visible and the pills looked more like buttons than tabs as well as are very blue (while the rest of the buttons in the UI are grey/black) so it it didn't really fit with the design IMO.

Otto-AA:
Do you think this has any advantages compared to the current version?
If not, I gonna close it

johan-bjareholt:
The advantages are that data-toggle is slightly nicer and that no CSS means less maintenance when bootstrap 5 is released, but since bootstrap 4 was just released maintenance shouldn't be an issue for a couple of years at least.

ErikBjare:
I don't think we'll merge this unfortunately, but thanks for attempting it!

Going forward, I'm interested in using Vuetify instead: https://github.com/ActivityWatch/aw-webui/issues/88

PR #85, opened 2018-06-26
Implemented audible-as-active

ErikBjare:
Very much WIP, very untested.

Depends on https://github.com/ActivityWatch/aw-core/pull/65


src/queries.js
```diff
@@ -1,11 +1,25 @@
 // TODO: Sanitize string input of buckets
 
+// Helper function returning a query that defines a variable events_active
+// that can be used to filter away non-active time.
+function events_active(afkbucket, audibleAsActive) {
+  return [
+    'events_active = [];',
```
johan-bjareholt: Why empty array and perdiod_union?
why not just `events_active = flood(query_bucket("${afkbucket}"));`?

src/queries.js
```diff
@@ -1,11 +1,25 @@
...
+    `events_afk = flood(query_bucket("${afkbucket}"));`,
+    'events_active = period_union(events_active, filter_keyvals(events_afk, "status", ["not-afk"]));',
+  ].concat(audibleAsActive ? [
+    'events_audible = filter_keyvals(events_browser, "audible", [true]);',
```
johan-bjareholt: Does `[true]` work?

src/queries.js
```diff
@@ -60,23 +71,20 @@ function editorActivityQuery (editorbucket, limit){
...
 }
 
 function dailyActivityQuery (afkbucket){
-  return [
-    'afkbucket = "' + afkbucket + '";',
-    'not_afk = flood(query_bucket(afkbucket));',
-    'not_afk = merge_events_by_keys(not_afk, ["status"]);',
-    'RETURN = not_afk;'
-  ];
+  return events_active().concat([
```
johan-bjareholt: This will make a huge performance hit, the aw-watcher-web bucket is much larger than aw-watcher-afk.
Might still be fast enough still though, I have no idea so you have to test.

src/views/Activity.vue
```diff
@@ -174,6 +174,8 @@ div
       div
         b-form-checkbox(v-model="filterAFK")
           | Filter away AFK time
+        b-form-checkbox(v-model="countAudibleAsActive")
```
johan-bjareholt: Nice!

src/queries.js
ErikBjare: Just experimented with a similar query in aw-research (https://github.com/ActivityWatch/aw-research/pull/12) and no, it doesn't.

johan-bjareholt: 1 or TRUE might work though, test that.

ErikBjare: Added boolean support in https://github.com/ActivityWatch/aw-core/pull/70/commits/3c4187e50bffb60a16593d671dbe4b301453e6e4

ErikBjare:
A complete (and working) version of the query was added to aw-research here: https://github.com/ActivityWatch/aw-research/pull/12

johan-bjareholt:
How's this PR doing?

Needs a rebase, but seems fine otherwise?

src/queries.js
ErikBjare: Ah shit, true. Don't know what I was thinking.

src/queries.js
```diff
@@ -1,11 +1,25 @@
...
+    `events_afk = flood(query_bucket("${afkbucket}"));`,
+    'events_active = period_union(events_active, filter_keyvals(events_afk, "status", ["not-afk"]));',
+  ].concat(audibleAsActive ? [
+    'events_audible = filter_keyvals(events_browser, "audible", [true]);',
+    'events_active = period_union(events_active, events_audible);',
+  ]);
+}
+
 function windowQuery(windowbucket, afkbucket, appcount, titlecount, filterAFK) {
+  // TODO: Take into account audible browser activity (tricky)
```
ErikBjare: This comment should probably be removed, after testing.

src/views/Activity.vue
ErikBjare: Thinking of removing this though, just adds unnecessary complexity. It's not like I'm expecting anyone to not want this, and it isn't saved across page reloads anyway which makes it kinda bad UX.

nicolae-stroncea:
@ErikBjare are there any blockers on this? I'm currently using some scripts to automatically send myself notifications based on total time spent on my PC and would love to take advantage of this feature :)

ErikBjare:
To merge this we need to first merge https://github.com/ActivityWatch/aw-webui/pull/203 and then update this PR to fit those changes.

nicolae-stroncea:
@ErikBjare could this be added to the 1.0 roadmap?

Currently absence of this PR  leads to wrong reporting any time you watch any media or have an in-browser call, and the reporting can be **way off**  depending on how long that activity goes for. Judging from number of issues on this spread through the AW repos, I would say a reasonable number of users see this as an important feature(since it directly affects the accuracy of the software)

johan-bjareholt:
@nicolae-stroncea I'm not sure what the 1.0 roadmap even means at this point, but I agree that it's an important change.

src/views/Activity.vue
johan-bjareholt: It is important if someone for example often leaves their computer for extended periods of time with for example Spotify or YouTube in the foreground while not actually being at the computer but only having it for playing music.

ErikBjare:
@johan-bjareholt The 1.0 roadmap is purely aspirational. I more or less made it to have an overview of more fundamental/longer-term issues, and "things we'd want in an eventual 1.0" label seemed like a good qualifier.

quentinus95:
It looks like all the explicit dependencies of this pull request have been merged. What is the remaining effort to get this feature available?

watertrainer:
What is the status of this? Does this still need testing (how could one test this)? I think this is easily one of the most important features for ActivityWatch, as this directly impacts, how accurate the data is.

src/queries.js
```diff
@@ -2,25 +2,46 @@ import _ from 'lodash';
 
 // TODO: Sanitize string input of buckets
 
+// Helper function returning a query that defines a variable events_active
+// that can be used to filter away non-active time.
+// TODO: Handle events from all available browser buckets, as done in: https://github.com/ActivityWatch/aw-webui/pull/108
+// TODO: Don't count audible as activity if window isn't active
```
ErikBjare, line 8: This is a blocker. 

I think we would need to finish implementing the combination of window events and browser events (https://github.com/ActivityWatch/aw-server-rust/pull/179) before making headway on it.

ErikBjare: Nevermind, it works just fine as I did it in #262

ErikBjare:
Closing in favor of #262

PR #92, opened 2018-08-02
Highlight Menu items when clicked

nicolae-stroncea:


johan-bjareholt:
Both me and @erikbjare are on vacation, will take a look when I have the time (hopefully on monday or tuesday)

src/App.vue
```diff
@@ -82,16 +82,24 @@ import 'vue-awesome/icons/search';
...
   data: function() {
     return {
       activity_hosts: [],
       connected: false,
+      current:"",
       info: {}
     }
   },
+  computed: {
+      isQueryPath: function(thepath) { return this.$route.path=="/query" },
```
johan-bjareholt: Order them in the same order as they are in the bar

src/App.vue
```diff
@@ -17,10 +17,10 @@ div#wrapper
   div.container.aw-container
     // TODO: Refactor into Mainmenu component
     b-nav.row.aw-navbar
-      b-nav-item(to="/")
+      b-nav-item(to="/" exact v-on:click="current='home'" :class="{ inCategory: current=='home' || isHomePath}"  )
         icon(name="home")
         | Home
-      b-nav-item(v-if="activity_hosts.length === 1", v-for="host in activity_hosts", :key="host", :to="'/activity/' + host")
+      b-nav-item(v-if="activity_hosts.length === 1", v-for="host in activity_hosts", :key="host", :to="'/activity/' + host" v-on:click="current='activity'" :class="{ inCategory: current=='activity' || isActivityPath}")
```
johan-bjareholt: This `v-on:clock="current='x'"` seems like it would be buggy to me.
What if you for example follow a link directly to activity without actually pressing the "Activity" button?
Why doesn't "isActivityPath" suffice?

nicolae-stroncea: "isActivityPath" does suffice. I posted the new commits for the updated version

ErikBjare:
@nicolae-stroncea Thanks for the contributions, I will merged your changes (with some modifications) in #100.

PR #100, opened 2018-08-26
Highlight menu items when clicked

ErikBjare:
Based on https://github.com/ActivityWatch/aw-webui/pull/92

Modifications:

 - Simplified menu highlight logic (see https://github.com/ActivityWatch/aw-webui/pull/100/commits/6e992122c3f6b8bf87f34511cbc3b08c172c5a7c)
    - Note how the class `active` is already automatically set). 
 - Reverted the "show only 5, show 5 less, show 5 more" buttons (I just don't like the look of it)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,4 +1,4 @@
-﻿<template lang="pug">
+<template lang="pug">
 div#wrapper
   div.header
     div.container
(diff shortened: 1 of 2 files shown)


nicolae-stroncea:
What if other icons are used rather than the ones in this PR?

A "show less" button is useful because sometimes I may want to go back and only see the top 5, or I find that the new data isn't that useful. Additionally, what currently ends up happening is if I switch to another day, I see the same amount of titles as I've selected to view on the previous day(i.e if i wanted to see top 30 items in top applications, i now have to view the same amount in all other days, as the counter doesn't reset[though this may be intended]). A "show 5" and "show by 5 less" solves those issues.

ErikBjare:
@nicolae-strongcea For now, the solution will be to simply reload the page. I don't want text-less buttons, and having 2 or more buttons with text is a bit much.

PR #106, opened 2018-08-29
Basic input component for time intervals

ErikBjare:
# TODO

 - [x] Make component return a moment-tuple `(start, end)` instead of a number of seconds.
 - [x] Rename component to input-timeinterval
 - [x] Style it properly


ErikBjare:
It's not that pretty, but it works really well.

PR #108, opened 2018-08-30
Combine and display chrome + firefox data in Browser Section

nicolae-stroncea:
Depends on https://github.com/ActivityWatch/aw-core/pull/70


ErikBjare:
@johan-bjareholt The best way to figure out if you've done flooding right is to simply inspect the events in the source buckets and the final events used in the query result (I'd suggest using the new timeline). 

But I think this is the right way to go, flooding should generally be done as early as possible.

nicolae-stroncea:
I tested both approaches initially. Flooding the window events led to the durations being too large, which I assume is due to both browser events intersecting with the same timestamps from the window watcher. Doing it separately ends up with the correct intersectiond

ErikBjare:
@nicolae-stroncea Alright, just fix the issues in https://github.com/ActivityWatch/aw-core/pull/70 and we'll merge this.

ErikBjare:
Also, it looks like you did something weird with `aw-client-js` which broke Travis. Check out the latest aw-client-js commit from master and commit that.

johan-bjareholt:
Oh, I forgot about this PR completely.

Fix the conflict and we'll merge this directly.

ErikBjare:
I've fixed the issues and cleaned up stuff in https://github.com/ActivityWatch/aw-webui/pull/141.

Closing this one, thanks for the amazing work @nicolae-stroncea! (and sorry for the delay)

PR #117, opened 2018-11-27
Fix bug where activity view doesn't use timestamp.

steventrouble:
The date() function returns the date in the URL directly, which doesn't use a timestamp. This means that anytime the app has a date in the URL, it will return results for the wrong timezone.

This changes the dateStart() method to use the ISO formatted date, which will always correctly specify a timezone.


ErikBjare:
Nice catch! Merging

PR #118, opened 2018-11-27
Fix git copypasta error.

steventrouble:
I am very embarrassed. I committed, then didn't amend it correctly.

As proof that this commit **does** compile and fix the bug, here's some screenshots:
Before: https://drive.google.com/open?id=1IsDohz3AU4Ie0NzTz3H4aMINiHVIVPF4
After: https://drive.google.com/open?id=1gSY0rh0uADF5iPF5EuWjOCDK3U9F8_rJ


ErikBjare:
Nice! Merging.

Fixes https://github.com/ActivityWatch/activitywatch/issues/254

PR #119, opened 2019-01-17
Added basic timer functionality

### LINKED ISSUE #70: Manual logging (timer/stopwatch)
Adding a timer functionality (a la Toggl) to the web UI would be really nice and pretty easy. 

The timer would be started by clicking a button next to a box used for labeling. This creates an event in a timer-bucket, with the additional label "timing" (or similar) for keeping track of unstopped timers.

The timer view should also have a way to edit start/end times of events as well as labels, would be nice to add this to the event-list view (so keep code reuse in mind).

ErikBjare:
Been looking for something simple to replace Toggl, was really easy.

### TODO

 - [x] Add ability to edit timers (start, stop, label)
     - Could be generalized into a general event editor
 - [ ] Add dropdown of recent labels for easier labelling

### What it looks like

image

Closes #70


src/views/Timer.vue
```diff
@@ -0,0 +1,107 @@
...
+          running: true,
+          label: this.label
+        }
+      }))
+    },
+
+    stopTimer: async function(event) {
+      event.data.running = false;
+      event.duration = (moment() - moment(event.timestamp)) / 1000;
+      await this.$aw.replaceEvent(this.bucket_id, event);
```
johan-bjareholt: Handle if the aw-server request fails? if that happens we should not have a stopped event in the webui.

src/views/Timer.vue
```diff
@@ -0,0 +1,107 @@
...
+    runningTimers() {
+      return _.filter(this.events, (e) => (e.data.running))
+    },
+    stoppedTimers() {
+      return _.filter(this.events, (e) => (!e.data.running))
+    }
+  },
+  methods: {
+    startTimer: async function() {
+      this.events.unshift(await this.$aw.heartbeat(this.bucket_id, 0, {
```
johan-bjareholt: unshift?

ErikBjare:
Looks pretty sick now IMO, even added a generalized event editor.

image

image

ErikBjare:
@johan-bjareholt I'm done with this PR now, sorry for the mess but it does a lot of neat things.

johan-bjareholt:
Just tried it out and realized that it isn't really what I expected.
I expected something like toggle where you can see start/endtime of each period, but each timer just seem to be a counter? If I start a timer, stop it and then resume it I expect there to be a gap between the first and the second intervals but it was all just a single long event. To be this seems kind of confusing because that means that the endtime is invalid if the event ever has been stopped+resumed.

johan-bjareholt:
What I'm trying to say is that I don't really understand the point of the resume button?
It would make sense to me to start a new timer with the same label at a later time, but not to restart the same timer and mess up the duration since the hole where it was stopped will be filled.

ErikBjare:
@johan-bjareholt Yeah, I kinda half-assed the resume-button. 

It probably makes more sense, as you say, to have a button that creates a new timer with the same label instead of a resume-button (which is really just a "I didn't actually mean to stop the timer"-button). I just now realize that this is what Toggl does.

johan-bjareholt:
Also, the timeline will probably look weird if we have two timers intersecting each other. I think that's fine though.

ErikBjare:
@johan-bjareholt Fixed the 'resume' to 'start new' thing. Merging.

johan-bjareholt:
@ErikBjare I don't think this is working properly yet and the UI is confusing.
It's fine that it's merged, but can we hide the "timer" button for now? I think we should think things through a bit more.

ErikBjare:
@johan-bjareholt There's a disclaimer at the top, is that not enough?

johan-bjareholt:
@ErikBjare Oh forgot about that. Yeah, that should be fine.

johan-bjareholt:
@ErikBjare As long as you are satisfied with the format of the bucket data there shouldn't be any issues, I have not looked very closely at that though.

src/views/Timer.vue
```diff
@@ -0,0 +1,105 @@
...
+  name: "Timer",
+  components: {
+    "timer-entry": TimerEntry
+  },
+  mounted: function() {
+    // TODO: List all possible timer buckets
+    //this.getBuckets();
+
+    // Create default timer bucket
+    this.$aw.ensureBucket(this.bucket_id, "timer", "unknown");
```
johan-bjareholt: We shouldn't use "timer" as a eventtype, we are trying to use eventtypes in the format of "os.window.current" etc.

johan-bjareholt: It's stuff like this that I don't want to be merged early, because once that's in and people start using it we can't change it

ErikBjare: Do you have a better suggestion for the event type? I can't really come up with anything.

johan-bjareholt: general.timer.stopwatch or something similar would be my suggestion
Timers are not really part of any os/web/whatever so general seems fine, it's a type of timer and the specific type of timer is a stopwatch.

ErikBjare: @karvla agreed. I went with `general.stopwatch`. See #130

Issue #120, opened 2019-01-18
Browser activity missing for firefox


vesatoivonen:
I'm on version v0.8.0b7 and using `aw-watcher-web-chrome` and `aw-watcher-web-firefox`.
Firefox bucket is always empty on activity tab, but on timeline I can see that firefox data is collected.
Chrome's data is shown propery on both views.

![screenshot_2019-01-18_11-09-34](https://user-images.githubusercontent.com/3157489/51377006-a5145080-1b01-11e9-93dc-d5f31b62f668.png)
![screenshot_2019-01-18_11-10-22](https://user-images.githubusercontent.com/3157489/51377148-0d633200-1b02-11e9-9637-a970aec28d03.png)

vesatoivonen:
I poked around a bit and it seems that this happens because I'm using Firefox Nightly and the filters used in the query don't match it.
Adding `FirefoxNightly` to `browser_appnames` on queries.js should fix this, at least for me.

johan-bjareholt:
For the web activity to be visible we also need aw-watcher-afk and aw-watcher-window to run at the same time. I can see in your timeline that aw-watcher-afk is working well (you have afk and not-afk events) but it seems like aw-watcher-window has not created any events in the timeline?
aw-watcher-window is needed so aw-server can know when the browser window is focused and not.

johan-bjareholt:
@vesatoivonen Well, that is also certainly a reason why it's not working!
Will you do a PR or should I add it to aw-webui?

vesatoivonen:
It would be great if you could add it, I'm not familiar enough with the project yet.
I confirmed by curling that the appname working for me is `FirefoxNightly`.
On mac and windows this might be something else.

johan-bjareholt:
Are you sure that the appname is exactly "FirefoxNightly"? No spaces or anything?
The other appnames we have seem to have spaces so I assume that this one might aswell, but it could also be a mistake on Firefox part.
These are the other appnames for firefox: `["Firefox", "Firefox.exe", "firefox", "firefox.exe", "Firefox Developer Edition", "Firefox Beta"]`

vesatoivonen:
Yeah, no spaces. This is a `eventmodel.datastr` straight from the db: `{"app": "FirefoxNightly", "title": "Preferences - Firefox Nightly"}`

ErikBjare:
[...]

PR #121, opened 2019-01-23
Include Firefox Nightly in the Firefox app names

punchagan:
Firefox Nightly's window name apparently is Nightly, which is weird. But, since we are only checking for events in the browser bucket, it should be fine to have such a generic name like Nightly?


johan-bjareholt:
That's odd as this comments says that it's "FirefoxNightly" https://github.com/ActivityWatch/aw-webui/issues/120#issuecomment-455553173

punchagan:
I'm not sure how exactly this window name gets set. I'm using builds from [here](https://launchpad.net/~ubuntu-mozilla-daily/+archive/ubuntu/ppa/+packages)

punchagan:
And similar to the other issue, adding 'Nightly' to the list fixed my dashboard.

johan-bjareholt:
Alright.

Since your code is tested with a version of Firefox Nightly from Mozillas official PPA I assume that yours is correct. If someone complains about nightly not working on aw-webui master later since they for some reason have a different app name we'll handle that then. Thanks for the PR!

punchagan:
Thanks for merging the PR, and the wonderful set of aw tools you are building!

PR #122, opened 2019-02-15
Hide header when webview detected

ErikBjare:
To make things look better on Android


johan-bjareholt:
This will not function properly if someone has open aw-server for the local network and you for example press a link to host:5600 in GMail and press it.

This use-case is extremely exotic though, so this change is fine.

ErikBjare:
@johan-bjareholt Oh right, hadn't considered that.

Apparently there is a way to manually set the User-Agent on a WebView. So if that use case becomes less exotic we could always just do that.

PR #128, opened 2019-02-27
Import and export buckets

johan-bjareholt:
Features:

-  Fixed so buckets properly exports with metadata
- Add "Import" and "Export all buckets" functionality to web-ui

Bugs:
- ~Neither Firefox nor Chrome seems to care about the download="aw-bucket-export_bucketname.json" and instead gets redirected to a new page with everything in plaintext and not caring about the filename in the "download" tag when pressing CTRL+S and simply gets named "export.json"~ Fixed in aw-server

### DIFF
--- a/src/views/Buckets.vue
+++ b/src/views/Buckets.vue
@@ -1,5 +1,10 @@
 <template lang="pug">
 div
+
+  b-alert(show)
+    | Are you looking to collect more data? Check out #[a(href="https://activitywatch.readthedocs.io/en/latest/watchers.html") the docs] for more watchers.
+    br
+    small #[b Note:] This is currently not as easy as we want it to be, so some familiarity with programming is currently needed to run most of them.
   h2 Buckets
 
   b-alert(variant="danger" :show="bucket_to_delete.length > 0")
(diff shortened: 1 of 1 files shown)


src/views/Buckets.vue
```diff
@@ -1,25 +1,25 @@
 <template lang="pug">
 div
+
+  b-alert(show)
+    | Are you looking to collect more data? Check out #[a(href="https://activitywatch.readthedocs.io/en/latest/watchers.html") the docs] for more watchers.
+    br
+    small #[b Note:] This is currently not as easy as we want it to be, so some familiarity with programming is currently needed to run most of them.
   h2 Buckets
 
   b-alert(variant="danger" :show="bucket_to_delete.length > 0")
```
ErikBjare: This should be moved into a modal, in another PR though.

johan-bjareholt:
@ErikBjare This and the aw-server dev/import code seems good to me now, review plz

src/views/Buckets.vue
```diff
@@ -28,19 +28,16 @@ div
...
           | Open bucket
       b-button-group(size="sm", class="mx-1")
-        // TODO: This currently does not export bucket metadata, which makes importing difficult
-        //       See: https://github.com/ActivityWatch/activitywatch/issues/103
-        // NOTE: When this is done we should also change the download name from "events-export" to "bucket-export".
-        b-button(:href="'/api/0/buckets/' + bucket.id + '/events?limit=-1'",
-                 :download="'aw-event-export-' + bucket.id + '.json'",
-                 title="Export all events from this bucket to JSON",
+        b-button(:href="$aw.baseURL + '/api/0/buckets/' + bucket.id + '/export'",
+                 :download="'aw-bucket-export-' + bucket.id + '.json'",
```
ErikBjare: Is the `:download` attribute even used at all?

PR #129, opened 2019-03-07
WIP: Sunburst view fix

johan-bjareholt:
Sunburst view doesn't work for me and doing this change fixes it.
The fix does not make sense, but it works.

For some reason when importing time in sunburst it does not contain seconds_to_duration, while on other modules where we also import time it does have seconds_to_duration.

### DIFF
--- a/src/visualizations/sunburst.js
+++ b/src/visualizations/sunburst.js
@@ -17,7 +17,7 @@
 const d3 = require("d3");
 const moment = require("moment");
 
-const time = require("../util/time.js");
+import time from "../util/time.js";
 const color = require("../util/color.js");
 
 // Dimensions of sunburst.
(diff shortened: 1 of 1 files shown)


ErikBjare:
I assume this has to do with using `require`. 

Try `import { seconds_to_duration } from '../util/time.js'` instead.

johan-bjareholt:
@ErikBjare Oh, sounds likely. Will test tonight.

This is the only bug that I could find that would be a blocker for next beta, if we get this to work we should to a release this weekend and do a new progress post

ErikBjare:
@johan-bjareholt I started writing a draft for the development update way back, you should be able to see it here: https://forum.activitywatch.net/t/development-update-february-2019/189

johan-bjareholt:
@ErikBjare Your guess was correct, merging

PR #135, opened 2019-03-16
Switched from (unmaintained) vis.js to timeline-plus

ErikBjare:
- Added removal of zero-duration events for performance reasons (removed a little over 1/3 of events during the 30min I worked on this)
 - Fixes scrolling on Firefox (and hopefully also Android)
 - Removes annoying special build step in Makefile to build vis.js
 - Also removed an old and unused Google Charts dependency

### DIFF
--- a/src/visualizations/VisTimeline.vue
+++ b/src/visualizations/VisTimeline.vue
@@ -39,6 +39,7 @@ export default {
   data () {
     return {
       timeline: null,
+      filterShortEvents: true,
       options: {
         zoomMin: 1000 * 60,             // 10min in milliseconds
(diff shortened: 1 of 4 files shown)


src/visualizations/VisTimeline.vue
```diff
@@ -130,7 +119,14 @@ export default {
         if(bucket.events === undefined) {
           return;
         }
-        _.each(_.sortBy(bucket.events, (e) => e.timestamp), (e) => {
+        let events = bucket.events;
+        // Filter out events shorter than 1 second (notably including 0-duration events)
+        // TODO: Use flooding instead, preferably with some additional method of removing/simplifying short events for even greater performance
```
johan-bjareholt: I was about to add a comment about this as well and then saw this comment

PR #137, opened 2019-04-10
Settings view and startofday setting

johan-bjareholt:
Was surprisingly easy to implement

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -8,6 +8,11 @@ import 'bootstrap/dist/css/bootstrap.css';
 import 'bootstrap-vue/dist/bootstrap-vue.css';
 Vue.use(BootstrapVue);
 
+// Setup default settings
+if (!("startOfDay" in localStorage)) {
+  localStorage.startOfDay = '04:00';
+}
+
 // Load the Varela Round font
 import 'typeface-varela-round';
 
--- a/src/views/Settings.vue
+++ b/src/views/Settings.vue
@@ -1,0 +1,42 @@
+
+<template lang="pug">
+div
+  h3 Settings
+
+  table.table
+    tbody
+      tr
+        td Start of day
+        td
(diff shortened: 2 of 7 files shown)


src/main.js
```diff
@@ -8,6 +8,15 @@ import 'bootstrap/dist/css/bootstrap.css';
 import 'bootstrap-vue/dist/bootstrap-vue.css';
 Vue.use(BootstrapVue);
 
+// Load and setup vue-cookies
+import VueCookies from 'vue-cookies'
```
ErikBjare: You should probably use localstorage instead.

src/main.js
```diff
@@ -8,6 +8,15 @@ import 'bootstrap/dist/css/bootstrap.css';
...
+Vue.use(VueCookies)
+// set default config
+VueCookies.config('aw-webui')
+if (!VueCookies.isKey('start-of-day')) {
+    VueCookies.set('start-of-day','02:00');
```
ErikBjare: I'd suggest a default of 04:00, that's what the Oura app uses (and some other apps I've used).

src/views/Settings.vue
```diff
@@ -0,0 +1,38 @@
...
+  h3 Settings
+
+  table.table
+    tbody
+      tr
+        td Start of day
+        td
+          input(type="time" :value="startOfDay" @change="setStartOfDay($event.target.value)")
+        td
+          | Sets the time where the start/end of a day is in the dayily activity view.
```
ErikBjare: Misspelling of 'daily'

src/main.js
johan-bjareholt: Fixed

src/main.js
johan-bjareholt: Fixed

src/views/Settings.vue
johan-bjareholt: Fixed

PR #141, opened 2019-05-07
Combine data from multiple browsers in Activity view

ErikBjare:
Continued from work by @nicolae-stroncea in https://github.com/ActivityWatch/aw-webui/pull/108


src/queries.js
```diff
@@ -64,30 +64,47 @@ export function appQuery(appbucket, limit) {
...
-  ] : [])
-  .concat([
-    'events = filter_period_intersect(events, window_browser);',
-    'events = split_url_events(events);',
+  // If multiple browser buckets were found
+  let code = (
+    `events = [];
+     window = flood(query_bucket("${windowbucket}"));`
+  ) + (
+    `not_afk = flood(query_bucket("${afkbucket}"));
```
johan-bjareholt: Could check for filter_afk here as well to avoid an unnecessary bucket query in that case.

ErikBjare: Done!

ErikBjare:
Merging now, thanks again to @nicolae-stroncea! :heart:

PR #145, opened 2019-09-21
Added classification vis & settings, switched to vue-cli, massiv…

ErikBjare:
### Now

 - [x] Fix issue with the summary not immediately showing when using/clicking the PeriodUsage bars to navigate (path needs `/summary` appended)
    - An option is to go back to the old architecture (without subroutes), which should be easy now that everything is refactored out and subroutes are decoupled from `$route.params`.
 - [x] Fix ErrorBoundary component (to avoid having alerts and error handlers in every component)
    - Sort of fixed. Works as long as all methods are properly `async`/`await`-
ed.
 - [x] Set the default classes in local storage, let user modify themselves in the settings.
 - [x] Finalize the category settings stuff ~~(adding/removing categories doesn't work right now, only editing)~~
 - [x] When parent category is renamed also rename eventual children

### After

 - [ ] Run `eslint --fix` on entire project in new PR that will be immediately merged.

### Later

 - [x] Add a category sunburst
 - [x] Start using Vuex properly
 - [ ] Merge web watcher events into window events in the query? (to allow for classifying by url/domain)
 - [ ] Category Tree: If a category has children, but also activity attributed directly to the parent that does not belong to a child, then create a "Other" child containing the activity.
 - [ ] Should we have a way to filter on a specific event key? Right now the regexes are applied on every (string) data value.
 - [x] Switch to Vue CLI 3 for easier config, better tooling, and future-proofing
 - [ ] Add a way to show how/which events are (un)categorized
 - [ ] Fix summary view to also use Vuex and all the new goodies
 

## Depends on
 - https://github.com/ActivityWatch/aw-core/pull/82
 - https://github.com/ActivityWatch/aw-server-rust/pull/67

## How it looks (for now)

image


johan-bjareholt:
There's a lot of nice changes here, but one primary thing missing here is the ability to do make your own tags and save them.

Could we focus on taking in the underlying aw-server changes here, then take in these aw-webui changes and then disable the view by default until we have solved that? I would hate to see this PR stagnate and need huge rebases in the future.

src/views/ActivityDaily.vue
```diff
@@ -385,13 +404,26 @@ export default {
...
+      let classes = [
+        ["Media -> Games", "RimWorld"],
+        ["Media -> Video", "YouTube"],
+        ["Media -> Social Media", "reddit|Facebook|Twitter"],
+        ["Remote", "Shadow"],
+        ["Work", "[Aa]lacritty|Google Docs"],
+        ["Work -> Programming", "[Pp]ython|GitHub"],
+        ["Work -> Programming -> ActivityWatch", "aw-|[Aa]ctivity[Ww]atch"],
+        ["Comms -> IM", "Messenger"],
+        ["Comms -> Email", "Gmail"],
```
johan-bjareholt: Don't use "->" and split the string, use arrays instead and pretty-print later with -> if that's what you want to do.

src/queries.js
```diff
@@ -26,31 +26,25 @@ export function summaryQuery(windowbucket, afkbucket, count) {
...
+    app_events   = sort_by_duration(merge_events_by_keys(title_events, ["app"]));
+    cat_events   = sort_by_duration(merge_events_by_keys(events, ["$category"]));
 
     events = sort_by_timestamp(events);
     app_chunks = chunk_events_by_key(events, "app");
     app_events  = limit_events(app_events, ${appcount});
     title_events  = limit_events(title_events, ${titlecount});
     duration = sum_durations(events);
-    RETURN  = {"app_events": app_events, "title_events": title_events, "app_chunks": app_chunks, "duration": duration};`;
+    RETURN  = {"app_events": app_events, "title_events": title_events, "cat_events": cat_events, "app_chunks": app_chunks, "duration": duration};`;
```
johan-bjareholt: As we have learned from earlier when we didn't have proper queries, sending all events to the web-ui and computing it there is not scalable for more than a days worth of data.
You should instead for example utilize the merge_events_by_keys for categories and write some new transform in the server for tags.

johan-bjareholt:
>  - [ ] Set the default classes in local storage, let user modify themselves in the settings.

We should not use browser storage for this, we should have an API in the server to save this.

We don't want all tags to be forgotten if we switch to a different browser.

ErikBjare:
>We should not use browser storage for this, we should have an API in the server to save this.
>We don't want all tags to be forgotten if we switch to a different browser.

Agreed, but just as with the next-day-start setting, I'm saying we could use browser storage until we have such a key/value store in aw-server for stuff like this.

src/queries.js
ErikBjare: But I am using `merge_events_by_keys`? I'm basically only sending one event per category in the `cat_events`.

ErikBjare:
>There's a lot of nice changes here, but one primary thing missing here is the ability to do make your own tags and save them.
>
>Could we focus on taking in the underlying aw-server changes here, then take in these aw-webui changes and then disable the view by default until we have solved that? I would hate to see this PR stagnate and need huge rebases in the future.

As I said, I think it'll be perfectly fine to use local storage until we have a way to store settings in aw-server.

I also want to refactor ActivityDaily.vue (using nested routes for the tabs) which I'll probably do in this PR. If I end up doing that I'll probably end up moving this stuff to a new tab with a "work in progress" disclaimer, which should prevent stagnation of this PR.

johan-bjareholt:
> As I said, I think it'll be perfectly fine to use local storage until we have a way to store settings in aw-server.

Alright, but could we move it out to a separate view so it's clear that it's WIP? I don't want to add WIP features onto the primary view both because it gets cluttered but also because we want it to be very clear that it's WIP.

> I also want to refactor ActivityDaily.vue (using nested routes for the tabs) which I'll probably do in this PR. If I end up doing that I'll probably end up moving this stuff to a new tab with a "work in progress" disclaimer, which should prevent stagnation of this PR.

I thought the plan was to refactor everything into the Summary view to then remove the daily view? Please don't waste time working on improving the daily view as that will just become duplicate work.

ErikBjare:
>I thought the plan was to refactor everything into the Summary view to then remove the daily view? Please don't waste time working on improving the daily view as that will just become duplicate work.

I had not heard of this? What makes the Summary view such a great replacement? I really don't see it. Anyway, before you replied I had already refactored the daily view massively, which should make it much easier to work with.

johan-bjareholt:
> I did not know this? What makes the Summary view such a great replacement? I really don't see it.

Mostly because of the fact that if you implement something once you get it for free for all other time ranges.
The original plan was to slowly move more and more features from the daily view once we found performant ways to do it in both so we afterwards only needed a single view.

johan-bjareholt:
> Mostly because of the fact that if you implement something once you get it for free for all other time ranges.
The original plan was to slowly move more and more features from the daily view once we found performant ways to do it in both so we afterwards only needed a single view.

Now that you have splitted out each section to its own file though that idea is probably moot as now each section is modular. Nice

johan-bjareholt:
Next time though please don't do massive linting like this in the same PR as actual code, it's really hard to read.

src/views/Settings.vue
```diff
@@ -3,42 +3,72 @@
...
   methods: {
     setStartOfDay: function(time_minutes) {
       localStorage.startOfDay = time_minutes;
-    }
+    },
+    addClass: function() {
+      this.classes.push({name: "New class", re: "FILL ME"})
+    },
+    saveClasses: function() {
+      saveClasses(this.classes);
```
johan-bjareholt: We should validate the regular expression before saving them. JS has built in re if i recall correctly.

ErikBjare:
> Next time though please don't do massive linting like this in the same PR as actual code, it's really hard to read.

Yeah, sorry about that.

> > Mostly because of the fact that if you implement something once you get it for free for all other time ranges.
> > The original plan was to slowly move more and more features from the daily view once we found performant ways to do it in both so we afterwards only needed a single view.
> 
> Now that you have splitted out each section to its own file though that idea is probably moot as now each section is modular. Nice

It will be even nicer when I get the whole thing into vuex and refactor the query builder.

src/views/Settings.vue
ErikBjare: The entire thing needs validation. This is probably not the final version of the UI anyway, but it works.

ErikBjare:
@johan-bjareholt It would be nice to have another review. The only thing missing is the ability to edit the categories/rules.

src/views/Dev.vue
```diff
@@ -0,0 +1,42 @@
...
+
+export default {
+  name: "aw-dev",
+
+  data: function() {
+    return {
+      queryOptions: {
+        aw_client: this.$aw,
+        date: moment().format('YYYY-MM-DD'),
+        host: 'erb-laptop2-arch',
```
johan-bjareholt: If you want this to go into master you should at least stop hardcoding this.
Just adding a text field or something as an input though should be good enough as i assume it's made to be a rough tool rather than for user debugging.

ErikBjare: Ah right, I forgot. Will do.

ErikBjare:
I got a sunburst working!

image

PR #146, opened 2019-10-14
Update queries.js, added Brave and Vivaldi browser

Matt-Soft:
Issue: no data display in Browser tab when using aw-watcher-web on Brave and Vivaldi browser on Archlinux.

Fix: 
On Archlinux Brave browser is detected by ua-parser-js as "chrome" so there is need to add "Brave-browser" in this group.


johan-bjareholt:
Looks good, thanks a lot!

PR #152, opened 2019-10-21
Testing Blockstack/Gaia for sync

ErikBjare:
I noticed apps like Nomie used Blockstack Gaia for sync, and it seems to work pretty well, so I wanted to give it a try.

I'm not sure if this will ever get merged. Much experimental. Very WIP.


ErikBjare:
Cool experiment, but I don't think this is happening. Closing.

PR #155, opened 2019-10-23
Update: improve README instructions

anthonkendel:
The README seemed outdated and the "Getting Started"
section didn't work.

In this commit a "Getting stared" section is added
to make it easier for other developers to setup
the project.

In addition to this some sections are moved/rewritten
to categorize and improve them under "Development" and
"Building".

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,41 +1,45 @@
-aw-webui
-========
+# aw-webui
 
 > A webui for ActivityWatch built in Vue.js
 
 [![Build Status](https://travis-ci.org/ActivityWatch/aw-webui.svg?branch=master)](https://travis-ci.org/ActivityWatch/aw-webui)
-[![Known Vulnerabilities](https://snyk.io/test/github/ActivityWatch/aw-webui/badge.svg)](https://snyk.io/test/github/ActivityWatch/aw-webui) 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,41 +1,53 @@
...
-## Build Setup
+```bash
+# init submodule
+git submodule init
 
-``` bash
-# Install dependencies
-make install
+# update submodule
+git submodule init
```
ErikBjare: These lines can be replaced with `git submodule update --init --recursive`

README.md
```diff
@@ -1,41 +1,53 @@
...
+
+# install aw-client-js
+cd aw-client-js
+npm install
+cd ..
+
+# install aw-webui dependencies
+npm install
```
ErikBjare: Both the aw-client-js stuff and npm install can be run in one go with `make install`, but since `make` isn't commonplace on macOS or Windows maybe this is better.

anthonkendel: I'm using WSL and `make install` didn't seem to work. I maybe missed something.

anthonkendel: Ok, so it didn't work because I didn't have the submodule updated. I'll update to `make install` to simplify the instructions.

anthonkendel:
> Awesome, thanks! Just a couple of minor things... :)

I think I fixed the things you request, but I can't seem to find them as I rewrite my commits.

PR #156, opened 2019-10-23
Fix: update project to ignore aw-client-js

anthonkendel:
The eslintIgnore that existed in package.json
did not work. This is moved to the .eslintignore
file and should work as intended in this commit.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 build/*.js
 config/*.js
+aw-client-js/**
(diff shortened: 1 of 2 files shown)


ErikBjare:
Hah, I must have missed that we already had a `.eslintignore` in https://github.com/ActivityWatch/aw-webui/commit/9fcb9161b7aaf669fd9dbcd4b2cf2b09bc9dacbc, thanks!

(Weird that the package.json thing didn't work for you though, works fine for me)

PR #157, opened 2019-11-05
Add 2hr option

timwsuqld:
1hr isn't enough, 4 hrs is too much, 2hr's is just right (for me). Thanks


ErikBjare:
The list is getting kinda long, but I'll allow it :)

timwsuqld:
Thanks @ErikBjare 
Maybe we need to make it something we can configure in the future, instead of hard coded.

PR #158, opened 2019-11-22
Fix for Vivaldi-snapshot browser logging on Linux

LeBaux:
Vivaldi has currently 2 branches, Vivaldi-stable and Vivaldi-snapshot. Added snapshot to the list. There is also Android version, and we might want to test logging on windows as well (eg. we might need vivaldi.exe added). This is potentially part of bigger issue, having the list of all browsers listed in appnames. Solves: https://github.com/ActivityWatch/activitywatch/issues/314#issuecomment-557589185

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -89,6 +89,7 @@ const appnames = {
   vivaldi: [
     'Vivaldi-stable',
     'Vivaldi-snapshot',
+    'vivaldi.exe',
   ],
 };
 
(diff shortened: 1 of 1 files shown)


ErikBjare:
As you mentioned [here](https://github.com/ActivityWatch/activitywatch/issues/314#issuecomment-557526871) it appears that Vivaldi stable is sometimes just `Vivaldi`. Maybe add that too?

LeBaux:
> As you mentioned here it appears that Vivaldi stable is sometimes just Vivaldi.

I double-checked, it looks it is indeed only `Vivaldi-stable` and `Vivaldi-snapshot` on Linux. I currently do not have access to a Windows machine, but I will have later this weekend. I will try to look into it.

ErikBjare:
I added `vivaldi.exe` myself, hopefully that should work on Windows. Merging.

PR #160, opened 2019-12-08
Add vuex support for editor activity

johan-bjareholt:
- Fixes editor support again so it works with other dates than today again (with vuex)
- Removed support for selecting a specific editor bucket
- Added back "duration" field for browser

Solves https://github.com/ActivityWatch/activitywatch/issues/316

### DIFF
--- a/src/store/modules/activity_daily.ts
+++ b/src/store/modules/activity_daily.ts
@@ -35,17 +35,25 @@ interface QueryOptions {
 const _state = {
   top_apps: [],
   top_titles: [],
+
+  browser_duration: [],
   top_domains: [],
   top_urls: [],
+
+  editor_duration: [],
+  top_editor_files: [],
+  top_editor_languages: [],
+  top_editor_projects: [],
+
   top_categories: [],
   active_events: [],
-  app_chunks: [],
-  web_chunks: [],
   active_duration: 0,
   active_history: {},
   query_options: {
     browser_buckets: 'all',
+    editor_buckets: 'all',
   },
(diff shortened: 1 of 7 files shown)


src/store/modules/activity_daily.ts
```diff
@@ -35,17 +35,26 @@ interface QueryOptions {
 const _state = {
   top_apps: [],
   top_titles: [],
+  app_chunks: [],
+
   top_domains: [],
   top_urls: [],
+  web_chunks: [],
```
ErikBjare: `app_chunks` and `web_chunks` aren't used anymore (because they make the query vastly more expensive).

johan-bjareholt: Done

ErikBjare:
I haven't tested, but looks good to me. Merging.

masaeedu:
Hello @johan-bjareholt. Do you know which version of aw-webui was this released in? Also, would you happen to know if there's a way I can check what version of aw-webui I'm running?

ErikBjare:
aw-webui is distributed with aw-server, which in turn is distributed with
the ActivityWatch bundle. The version you're running should be visible in
the bottom right corner of the web UI.

Whichever release was made after the PR was merged should contain the fix,
if none such exists, it will be in the next one.

johan-bjareholt:
> Whichever release was made after the PR was merged should contain the fix,
if none such exists, it will be in the next one.

@masaeedu  It is not in a release yet, so next version will have it!

PR #161, opened 2019-12-16
Implemented tagging ability in Stopwatch

nicolae-stroncea:
User can write a list of tags for his stopwatch event that will be automatically stored inside the database. This functionality is useful for data-analysis, where you might want to view your stopwatch events by different tags: study, work, etc.

Screenshot from 2019-12-16 14-45-22


ErikBjare:
Nice! I've been thinking about building something similar myself recently, have you thought about how this could work together with categorization?

johan-bjareholt:
@ErikBjare We could add stopwatch events as a separate tab similar to "browser" and "editor" and then just straight up run the categorization rules there.

That in addition to manually add stopwatch events would be a pretty decent way to manually track time.

ErikBjare:
@johan-bjareholt That's a good start. 

In my notebooks where I build a 'canonical timeline' I use stopwatch-like events (from Toggl) as filler when no other event exists. So that if there is no activity for a 20min period, but there's a stopwatch event, the stopwatch event gets slotted in. Basically it gets a lower selection priority than automatically tracked time, which is just what I assume most people want since it's effectively an 'AFK timer'.

nicolae-stroncea:
I was thinking that additionally to a stopwatch bucket, there can be a general bucket for **user-events**(a period of time the user annotates to explain what he's doing in that time, exactly like stopwatch). The user may want to split the day into things he has done, without running a timer on them, especially if he's afk. Stopwatch events could be part of this bucket, since they're also user-annotated events, but I think adding them in the same bucket might be cleaner. I think most of the functionality for this is already built into the StopWatch but a more general bucket might be helpful(or just renaming it since it could include any user-annotated events).

Here are some ideas I had on doing analysis on user-events from that bucket:

##### Creating customized query
* The user can create his own query to analyze user-events. He can select the user-event attributes that he wants to query for:
  * By ID: select the user-event which has a specific ID.
  * By Name: select all user-events which have a specific name
  * By Tags: select all user-events which have a tag from a list.
  * By time period: you select all user-events in a time period.
  * On the front-end this can be done with some key-value pairs, exactly like the Edit Event.
* The backend then goes through the bucket, and generates a list of TimePeriods based on all user-events which fit the criteria.
* It then generates the same statistics and visualizations as the summary page, but specifically for the timeperiods the user selected. I think all API calls currently support lists of timeperiods.
* **Applications**: This way the user can then see: from all the time periods he annotated as studying how much studying did he actually do? Or what kinds of activities did he do in a specific timeslot(if he annotated it as Programming, how much time did he actually spend on social media)?

##### Default Query

* The default query can be selecting all the user-events from Today

src/components/StopwatchEntry.vue
```diff
@@ -54,6 +54,14 @@ export default {
     stop: async function() {
       this.event.data.running = false;
       this.event.duration = (moment() - moment(this.event.timestamp)) / 1000;
+      // Convert the string to a list of tags
+      if(this.event.data.tags){
+        this.event.data.tags = this.event.data.tags.replace(/ /g,'')
+        this.event.data.tags = this.event.data.tags.split(',')
```
johan-bjareholt, line 60: This only allows for a single word in the tag, multi-word tags will be destroyed.
Would be better if it finds a comma it should then removes all types of spaces before it and after it.

nicolae-stroncea:
Sorry for the late reply haha. I fixed the single-word issue, and merged most recent changes from Stopwatch.

Let me know if anything else is needed to merge this PR

nicolae-stroncea:
Realized there's some mistakes in the code, fixing and will commit again soon

nicolae-stroncea:
Created a new PR instead, closing this one

PR #162, opened 2019-12-17
Annotating Productivity and Visualizing

nicolae-stroncea:
* User can select between: Productive, Not Productive, Neutral. Default is Unknown.
* The categories are then aggregated, which enables the user to view how productive he was during a certain period.
Screenshot from 2019-12-17 03-58-35

Screenshot from 2019-12-17 03-58-27
Screenshot from 2019-12-17 03-58-16

### Known Issues
* Will work in a completely new environment(or if defaults are reset), but there are currently bugs if you start it in an environment where Productivity has not yet been set for already existing categories.

### DIFF
--- a/src/components/CategoryEditTree.vue
+++ b/src/components/CategoryEditTree.vue
@@ -8,8 +8,11 @@ div
     div.col-4.col-md-8
       //span.d-none.d-sm-inline.d-md-none(:style="{ marginLeft: (2 * depth) + 'em'}")
       span.d-none.d-md-inline
-        span(v-if="cls.rule.type === 'regex'") Rule ({{cls.rule.type}}): #[code {{cls.rule.regex}}]
-        span(v-else, style="color: #888") No rule
+        span(v-if="cls.rule.type === 'regex'") Rule ({{cls.rule.type}}): #[code {{cls.rule.regex}}], 
+        span(v-else, style="color: #888") No rule, 
+      span.d-none.d-md-inline
+        span(v-if="cls.productivity !== -1") {{ productivityType }}
+        span(v-else, style="color: #888") Unknown Productivity
       span.float-right
         b-btn.ml-1(size="sm", variant="outline-secondary", @click="showEditModal($event)" style="border: 0;" pill)
           icon(name="edit")
@@ -25,7 +28,9 @@ div
         b-form-input(v-model="editing.name")
       b-input-group(prepend="Parent")
         b-select(v-model="editing.parent", :options="allCategories")
-
+    div.my-1
+      b-input-group.mv-1(prepend="Productivity")
+        b-select(v-model="editing.productivity", :options="allProductivityTypes")
     hr
 
     div.my-1
@@ -70,8 +75,15 @@ export default {
         id: 0, // FIXME: Use ID assigned to category in vuex store, in order for saves to be uniquely targeted
         name: null,
         rule: {},
+        productivity: -1,
         parent: [],
       },
+      productivity_description:{
+        '-1':"Unknown Productivity",
+        '0':"Not Productive",
+        '1': "Productive",
+        '2': "Neutral Productivity"  
+      }
     };
   },
   computed: {
@@ -87,10 +99,21 @@ export default {
         //{ value: 'glob', text: 'Glob pattern' },
       ]
     },
+    productivityType: function(){
+      return this.productivity_description[this.cls.productivity]
+    },
+    allProductivityTypes: function(){
+      return [
+        { value: -1,text:this.productivity_description[-1]},
+        { value: 0, text:this.productivity_description[0]},
+        { value: 1, text:this.productivity_description[1]},
+        { value: 2, text:this.productivity_description[2]}
+      ]
+    }
   },
   methods: {
     addSubclass: function(parent) {
-      this.$store.commit('settings/addClass', {name: parent.name.concat(["New class"]), rule: {type: "regex", regex: "FILL ME"}});
+      this.$store.commit('settings/addClass', {name: parent.name.concat(["New class"]), productivity: parent.productivity, rule: {type: "regex", regex: "FILL ME"}});
     },
     removeClass: function(cls) {
       // TODO: Show a confirmation dialog
@@ -122,6 +145,7 @@ export default {
         id: this.editing.id,
         name: this.editing.parent.concat(this.editing.name),
         rule: this.editing.rule.type !== null ? this.editing.rule : { type: null },
+        productivity: this.editing.productivity ? this.editing.productivity : -1
       };
       this.$store.commit("settings/updateClass", new_class);
 
@@ -135,6 +159,7 @@ export default {
         id: this.cls.id,
         name: this.cls.subname,
         rule: _.cloneDeep(this.cls.rule),
+        productivity: this.cls.productivity ? this.cls.productivity : -1,
         parent: this.cls.parent ? this.cls.parent : [],
       };
       //console.log(this.editing);
--- a/src/components/CategoryEditTree.vue
+++ b/src/components/CategoryEditTree.vue
@@ -11,7 +11,7 @@ div
         span(v-if="cls.rule.type === 'regex'") Rule ({{cls.rule.type}}): #[code {{cls.rule.regex}}], 
         span(v-else, style="color: #888") No rule, 
       span.d-none.d-md-inline
-        span(v-if="cls.productivity !== -1") {{ productivityType }}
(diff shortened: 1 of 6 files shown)


johan-bjareholt:
> Will work in a completely new environment(or if defaults are reset), but there are currently bugs if you start it in an environment where Productivity has not yet been set for already existing categories.

Should be simple to fix right? Assuming "Unknown Productivity" if nothing is set

src/components/CategoryEditTree.vue
```diff
@@ -70,8 +75,15 @@ export default {
         id: 0, // FIXME: Use ID assigned to category in vuex store, in order for saves to be uniquely targeted
         name: null,
         rule: {},
+        productivity: -1,
         parent: [],
       },
+      productivity_description:{
```
johan-bjareholt: I don't think this is a good way to do it.

Don't ever set "Unknown Productivity", in that case just don't set the productivity for that rule and make the transform skip that rule.

Then for example set 0 as neutral, 1 as productive and -1 as unproductive. Then you could for example later give the productivity score for a day/week/month/year a scale from -1 to 1

nicolae-stroncea: @johan-bjareholt 
are you saying to completely remove the attribute if it's never set? If we don't set it to anything, then some categories will have a productive attribute, some won't, which might complicate the checking that needs to be done when we're parsing events. What do you think about flipping it instead to: "not productive":-1, "neutral":0, "productive":1, "unknown":2.

ErikBjare:
This is really nice work, but I kinda had this use-case in mind when I designed categorization in mind. So I already have some opinions on how I think this should be done, and it's somewhat different.

When I built the categorization stuff, the intention was for the user to be able to create *multiple sets of categories*. One is the "user set" that we have today, one could be an open/standardized set, and one could be a "Productive/Unproductive" set.

A query would be something like:

```python
events = get_events()
events = categorize(events, cats_user)
events_productive = merge_events_by_key(categorize(events, cats_productivity), "$category")
```

This has some benefits:

 - The current user-defined set of categories will not match perfectly to productivity levels, this way of doing it gives more flexibility, and doesn't require introducing new object attributes.
 - No new change to server code needed (it was by design), would already work with both aw-server and aw-server-rust.
 - No built-in assumptions in the API about what labels like "productive" and "unproductive" means in value terms (so not a 1 for "productive" and -1 for unproductive, that's an analysis step). I really want to avoid building in assumptions like this in the APIs (and I think @johan-bjareholt will agree). This always annoyed me with RescueTime which calculated the daily productivity score in a really dumb way, (basically: `100 * (productive_time - unproductive_time) / total`) which I could write a whole rant about (basically it incentivizes not being on the computer at all if you've already got a little bit of productive time without any unproductive time, it's only natural to slack as you get tired, what's important is the amount of work done).

Note that you can still categorize productivity using previous categorization data, since "$category" is just an event attribute.

It makes rules applicable to categories, instead of making specific categories applicable to rules (which is what this way of doing it does).

What do you think @nicolae-stroncea and @johan-bjareholt? Sorry I didn't really write about these plans anywhere before you put a bunch of work in it, but I could build it on top of these changes so it comes to use and you get your credit :slightly_smiling_face:

nicolae-stroncea:
> This is really nice work, but I kinda had this use-case in mind when I designed categorization. So I already have some opinions on how I think this should be done, and it's somewhat different.
> 
> When I built the categorization stuff, the intention was for the user to be able to create _multiple sets of categories_. One is the "user set" that we have today, one could be an open/standardized set, and one could be a "Productive/Unproductive" set.
> 
> A query would be something like:
> 
> ```python
> events = get_events()
> events = categorize(events, cats_user)
> events_productive = merge_events_by_key(categorize(events, cats_productivity), "$category")
> ```
> 
> This has some benefits:
> 
>     * The current user-defined set of categories will not match perfectly to productivity levels, this way of doing it gives more flexibility, and doesn't require introducing new object attributes.
> 
>     * No new change to server code needed (it was by design), would already work with both aw-server and aw-server-rust.
> 
>     * No built-in assumptions in the API about what labels like "productive" and "unproductive" means in value terms (so not a 1 for "productive" and -1 for unproductive, that's an analysis step). I really want to avoid building in assumptions like this in the APIs (and I think @johan-bjareholt will agree). This always annoyed me with RescueTime which calculated the daily productivity score in a really dumb way, (basically: `100 * (productive_time - unproductive_time) / total`) which I could write a whole rant about (basically it incentivizes not being on the computer at all if you've already got a little bit of productive time without any unproductive time, it's only natural to slack as you get tired, what's important is the amount of work done).
> 
> 
> Note that you can still categorize productivity using previous categorization data, since "$category" is just an event attribute which will be used in categorization.
> 
> It makes rules applicable to categories, instead of making specific categories applicable to rules (which is what this way of doing it does).
> 
> What do you think @nicolae-stroncea and @johan-bjareholt? Sorry I didn't really write about these plans anywhere before you put a bunch of work in it, but I could build it on top of these changes so it comes to use and you get your credit slightly_smiling_face

@ErikBjare  not a problem, I should have probably asked as well to confirm whether there was anything in the works and what the general ideas were on the implementation. The commits are now in the PR's for aw-webui and aw-core, so feel free to take the bits and pieces that fit with your implementation.

johan-bjareholt:
> No built-in assumptions in the API about what labels like "productive" and "unproductive" means in value terms (so not a 1 for "productive" and -1 for unproductive, that's an analysis step). I really want to avoid building in assumptions like this in the APIs (and I think @johan-bjareholt will agree). 

Agreed, however if you could weigh it as 1.0 for "very productive", 0.5 for "productive" and -1 for "very unproductive" it would be an acceptable start as we could in the future upgrade that to a smarter algorithm for weighing the productivity.

> This always annoyed me with RescueTime which calculated the daily productivity score in a really dumb way, (basically: 100 * (productive_time - unproductive_time) / total) which I could write a whole rant about (basically it incentivizes not being on the computer at all if you've already got a little bit of productive time without any unproductive time, it's only natural to slack as you get tired, what's important is the amount of work done).

I agree that it's stupid that it's so black-and-white (either productive or unproductive and nothing in between). Otherwise I don't see it as a flaw that it's better to leave the computer to avoid slacking at the computer even though your AFK time could potentially be unproductive. Do you have any more reasons?

src/components/CategoryEditTree.vue
johan-bjareholt: >  What do you think about flipping it instead to: "not productive":-1, "neutral":0, "productive":1, "unknown":2.

See my comment below about I'd prefer if it was a range from -1 to 1 rather than static enums.

ErikBjare:
I just want to add here that https://github.com/ActivityWatch/activitywatch/issues/416 would work really well with my proposal for how this feature should be built (multiple category-sets).

ErikBjare:
Hey @nicolae-stroncea, I just wanted to tell you how much I appreciate your work on this PR even though it might not be merged.

I'm still not convinced of the best/cleanest way to implement this, but this PR really got the gears turning and I'll try to salvage whatever I can from this PR when we eventually get around to it, so you get your commit-credit where due :slightly_smiling_face:

nicolae-stroncea:
@ErikBjare  I really like the idea of scoring each category. This gives the user full flexibility of choosing whether they want the standard Productive:1, Unproductive:-1, Neutral:0, or something more powerful that is custom to their preferences.

Here's a possible approach:

Implementation:

1. Every category has a *productivity score* field that is stored along with the category. Therefore, it is currently stored in localStorage, and in the future will be stored in the db.
2. *productivity score* is any *whole* number(+ve or -ve). 
2. A newly created category has *productivity score* 0, and newly created subcategory has *productivity score* of its parent.

New views
1. View from this PR: horizontal bar chart  of each different productivity score and the total amount of time on it.
2. Productivity Score Number = *productivity score* x  *total time with this score* / *total time*

nicolae-stroncea:
@ErikBjare if above implementation makes sense to you and @johan-bjareholt , I can modify this PR accordingly. I'll be busy for a couple of weeks but I can give this a shot once I free up some time.

ErikBjare:
@nicolae-stroncea I've rebased this PR on master and adjusted a few things in #258 which makes the implementation a lot simpler (no need to add a transform to the backend). 

Once I realized the similarities between the 'productivity score' and the category colors I implemented in #254 it became pretty clear how this should be implemented, and I anticipate we can get it merged pretty soon!

I appreciate you might not have much time in the coming week, but I'd be really grateful if you had a look and reviewed it :)

nicolae-stroncea:
Closing this since #258  was merged

PR #163, opened 2019-12-20
Fix firefox dev edition browser data not showing up in activity view

chrsep:
Fixes ActivityWatch/activitywatch#324. Tested and works on my computer.


johan-bjareholt:
Thanks! :)

PR #165, opened 2019-12-29
stopwatch: Fixes broken "stop" and "edit" 

johan-bjareholt:
There were three bugs:
1. When pressing "stop" we modified the event within StopwatchEntry which
   tried to modify the event. Since javascript is reference by value the
   actual event in the list was never modified, only the copy. This created
   lots of Vue errors every time you pressed "stop"
2. When editing an even it was the same thing, the action was saved in aw-server
   but it was not shown in the webui since the parent component was not aware of
   the change so you have to reload the webpage.
3. We were using heartbeats instead of an insert when creating events which do
   not return anything in aw-server-rust or in aw-server-python with the sqlite
   backend. This meant that id was always "null" so instead of stopping the
   timer a new identical was created (unless you refreshed the page, then the
   usual getEvents was used so the id was correct).

### DIFF
--- a/src/components/EventEditor.vue
+++ b/src/components/EventEditor.vue
@@ -33,9 +33,9 @@ div
   hr
 
   div.float-right
-    b-button.mx-1(@click="$emit('cancel')")
+    b-button.mx-1(@click="$emit('close');")
       | Cancel
-    b-button.mx-1(@click="save()", variant="primary")
+    b-button.mx-1(@click="save(); $emit('close');", variant="primary")
       | Save
 </template>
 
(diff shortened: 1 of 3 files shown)


johan-bjareholt:
@ErikBjare Works well now, please review!

ErikBjare:
Not really related to this PR, but we should probably move the state to a Vuex module before we improve the view further.

johan-bjareholt:
> Not really related to this PR, but we should probably move the state to a Vuex module before we improve the view further.

For Stopwatch and StopwatchEntry I agree, for EventEditor though I think this is a better solution with events since then we could re-use that for editing events in other places as well.

PR #166, opened 2020-01-13
Improvements to Stopwatch and EventEditor

ErikBjare:
- [x] Stopwatch now shows all timers in a list with sections by date (instead of two columns divided by running/stopped)
   - Looks a lot better and is basically how Toggl does it
 - [x] Added ability to edit start/end of event
   - ~~Not very well tested~~
 - [x] Added ability to open EventEditor from Bucket view
   - ~~Known to not work~~

Adds dependencies needed for vue-datetime.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,13 +102,16 @@
     "jsdom": "^13.2.0",
     "less": "^3.0.4",
     "less-loader": "^4.1.0",
+    "luxon": "^1.21.3",
     "style-loader": "^0.21.0",
     "timeline-plus": "^2.3.6",
     "typeface-varela-round": "0.0.54",
     "vue-awesome": "^3.5.1",
     "vue-chartjs": "^3.4.2",
     "vue-d3-sunburst": "^1.0.0",
+    "vue-datetime": "^1.0.0-beta.11",
     "vue-property-decorator": "^8.3.0",
-    "vuex": "^3.1.1"
+    "vuex": "^3.1.1",
+    "weekstart": "^1.0.1"
   }
 }
--- a/src/components/EventEditor.vue
+++ b/src/components/EventEditor.vue
@@ -84,13 +84,6 @@ export default {
       await this.$aw.replaceEvent(this.bucket_id, this.editedEvent);
       this.$emit('save', this.editedEvent);
     },
-    endformat(end) {
-      if (moment(this.event.timestamp).isSame(end, 'day')) {
-        return "HH:MM";
-      } else {
-        return end;
-      }
-    }
   }
 }
 </script>
--- a/src/components/EventEditor.vue
+++ b/src/components/EventEditor.vue
@@ -63,10 +63,9 @@ export default {
         return moment(this.editedEvent.timestamp).format();
       },
       set: function(dt) {
-        const start = moment(dt);
-        const end = moment(this.end);
-        this.editedEvent.timestamp = start.format();
-        this.editedEvent.duration = end.diff(start, 'seconds');
+        // Duration needs to be set first since otherwise the computed for end will use the new timestamp
+        this.editedEvent.duration = moment(this.end).diff(dt, 'seconds');
+        this.editedEvent.timestamp = new Date(dt);
       }
     },
     end: {
@@ -81,8 +80,9 @@ export default {
   },
   methods: {
     async save() {
-      await this.$aw.replaceEvent(this.bucket_id, this.editedEvent);
+      // This emit needs to be called first, otherwise it won't occur for some reason
       this.$emit('save', this.editedEvent);
+      await this.$aw.replaceEvent(this.bucket_id, this.editedEvent);
     },
   }
 }
--- a/src/components/EventEditor.vue
+++ b/src/components/EventEditor.vue
@@ -9,13 +9,13 @@ div
       td {{ event.id }}
     tr
       th Start
-      td {{ event.timestamp | iso8601 }}
+      datetime(type="datetime" v-model="start")
     tr
       th End
-      td {{ 1 * event.timestamp + 1000 * event.duration | iso8601 }}
+      datetime(type="datetime" v-model="end")
     tr
       th Duration
-      td {{ event.duration | friendlyduration }}
+      td {{ editedEvent.duration | friendlyduration }}
 
   hr
 
@@ -43,6 +43,9 @@ div
 </style>
 
 <script>
+import moment from 'moment';
+
+
 export default {
   name: "EventEditor",
   props: {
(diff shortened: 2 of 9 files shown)


ErikBjare:
@johan-bjareholt Ready for review!

package.json
```diff
@@ -102,13 +102,16 @@
...
     "timeline-plus": "^2.3.6",
     "typeface-varela-round": "0.0.54",
     "vue-awesome": "^3.5.1",
     "vue-chartjs": "^3.4.2",
     "vue-d3-sunburst": "^1.0.0",
+    "vue-datetime": "^1.0.0-beta.11",
     "vue-property-decorator": "^8.3.0",
-    "vuex": "^3.1.1"
+    "vuex": "^3.1.1",
+    "weekstart": "^1.0.1"
```
johan-bjareholt: Can't see where this dependency is used?

src/visualizations/EventList.vue
```diff
@@ -25,6 +25,13 @@ div
               icon(name="tags")
               // TODO: Add some kind of highlighting to key
               | {{ key }}: {{ val }}
+            span.field.float-right
+              b-btn(v-b-modal="'edit-modal-' + event.id", variant="outline-dark" size="sm")
+                | Edit
+
+          // TODO: Refactor modal into event-editor component?
```
johan-bjareholt: Good idea!

src/views/Bucket.vue
```diff
@@ -63,10 +64,19 @@ export default {
...
       });
+      this.events = this.bucket_with_events.events;
     },
     getEventCount: async function(bucket_id) {
       this.eventcount = (await this.$aw.countEvents(bucket_id)).data;
     },
+    updateEvent: function(event) {
+      const i = this.events.findIndex((e) => e.id == event.id);
+      if(i != -1) {
+        this.$set(this.events, i, event);
```
johan-bjareholt: Would be nice with a comment like I did in my previous change why this.$set is needed and just modifying this.events is not sufficient. It's an detail that is easy to miss otherwise.

src/components/StopwatchEntry.vue
```diff
@@ -1,9 +1,12 @@
...
 div
-  div.row.p-3#root
+  div.row.px-3.py-2#root
     div.flex-fill
-      div #[b {{event.data.label || 'No label'}}]
-      div
+      span #[b {{event.data.label || 'No label'}}]
+      span(style="color: #888") &nbsp;|&nbsp;
+      span(v-if="event.data.running")
+        | Running for #[span(:title="event.timestamp") {{event.data.running ? (now - event.timestamp) / 1000 : event.duration | friendlyduration}}] (Started {{ event.timestamp | shorttime }})
```
johan-bjareholt: This is an insanely long line...

src/components/StopwatchEntry.vue
```diff
@@ -1,9 +1,12 @@
...
+      span(v-else)
         | Started #[span(:title="event.timestamp") {{event.timestamp | friendlytime}}] ({{event.data.running ? (now - event.timestamp) / 1000 : event.duration | friendlyduration}})
```
johan-bjareholt: This line is also insanely long, assume that's where you got it from?

src/components/EventEditor.vue
```diff
@@ -9,13 +9,13 @@ div
       td {{ event.id }}
     tr
       th Start
-      td {{ event.timestamp | iso8601 }}
+      datetime(type="datetime" v-model="start")
```
johan-bjareholt: This selector is kind of strange to me, very fancy but still very inconvenient (a big calendar, touch optimized scroll wheel clock on a desktop and without the ability to select seconds).
In my opinion it's not much better than just having a string for the date and time with proper validation (but I guess proper validation is very of hard to do).

ErikBjare: I looked around for good datetime components, and this was the best I could find. I agree with your concerns, but I do think it is better than a plain string with validation.

I looked at how Toggl does it, and I considered splitting them into different input fields (one for date, one for time) for easier editing, but that would be a lot of extra work.

package.json
ErikBjare: It's a peer dependency of the datetime picker.

src/visualizations/EventList.vue
ErikBjare: Wasn't sure how to do this at first glance however, due to how modals are enabled (see the v-b-modal="blabla" a couple of lines above). So post-poned for now.

src/views/Stopwatch.vue
```diff
@@ -73,7 +76,10 @@ export default {
     },
     stoppedTimers() {
       return _.filter(this.events, (e) => (!e.data.running))
-    }
+    },
+    timersByDate() {
```
johan-bjareholt: Don't you also need to sort them as well?
When I edit a timer with a date that is older than the rest it always shows up in the top, then when I refresh the page it moves to the bottom as expected.
Should be fixed somehow.

ErikBjare: Ah yes, nice find. Need to v-for over the sorted keys.

johan-bjareholt:
We should also be VERY careful when modifying dates as aw-server-python accepts any string as a date which can break the database completely in all datasets where this invalid date is present. So for example if a single event in an aw-watcher-window bucket is malformatted or invalid it will break the Activity view for where that event is present in both the day, week, month and year view.

aw-server-rust properly validates dates so no wrongly formatted or invalid dates are not accepted very early in the parsing stage when the request gets parsed from the webserver.

src/views/Stopwatch.vue
```diff
@@ -17,22 +17,25 @@ div
...
-        stopwatch-entry(:event="e", :bucket_id="bucket_id", :now="now",
-          @delete="deleteTimer", @update="updateTimer", @new="startTimer(e.data.label)")
-        hr(style="margin: 0")
+      div.mb-2
+
+      div(v-for="(v, k) in timersByDate")
+        h5.mt-4 {{ k }}
+        div(v-for="e in timersByDate[k]")
+          stopwatch-entry(:event="e", :bucket_id="bucket_id", :now="now",
+            @delete="deleteTimer", @update="updateTimer", @new="startTimer(e.data.label)")
```
johan-bjareholt: I'm not sure if I broke it or if you broke it but "deleteTimer" seems to be broken

src/components/EventEditor.vue
johan-bjareholt: We can make an github issue for it, it's no hurry. Should work fine for now.

ErikBjare:
>aw-server-python accepts any string as a date which can break the database completely in all datasets where this invalid date is present.

This seems completely crazy. Really?

I tried this:

```
curl -H "Content-Type: application/json" -d '{"timestamp": "asd", "data": {"a": "asd"}}' localhost:5666/api/0/buckets/aw-stopwatch/events
```

And it failed fine: `iso8601.iso8601.ParseError: Unable to parse date string 'asd'`

src/views/Stopwatch.vue
ErikBjare: Fixed!

ErikBjare:
Found a bunch of other bugs, they should all be fixed now.

johan-bjareholt:
> > aw-server-python accepts any string as a date which can break the database completely in all datasets where this invalid date is present.
> 
> This seems completely crazy. Really?
> 
> I tried this:
> 
> ```shell
> curl -H "Content-Type: application/json" -d '{"timestamp": "asd", "data": {"a": "asd"}}' localhost:5666/api/0/buckets/aw-stopwatch/events
> ```
> 
> And it failed fine: `iso8601.iso8601.ParseError: Unable to parse date string 'asd'`

I don't remember how, but this has happened to me before. I had to use a sqlite browser program to open the database, find the corrupt event and then delete it manually and start aw-server again.


Will take a look at the changes tonight.

PR #167, opened 2020-01-16
More software to categorize

Mte90:
I just saw that ActivityWatch doesn't categorize some of the software/sites that I use, so I just thought that can be useful for the whole project.


ErikBjare:
Hey, thanks for the PR! I realize that we need to improve the default categories, and want to make sure we do it correctly to minimize the possibility of miscategorization.

I'll open an issue where we can accept and discuss suggestions for additions to the default categories to achieve the best possible defaults :)

Mte90:
Any updates for this? In the meantime is possible to add them?

Mte90:
Updated

johan-bjareholt:
Better to take them in than doing nothing.
The performance should be about the same since the amount of rules are still the same.
Merging.

PR #169, opened 2020-01-30
Added setting for 'show last' duration default value

riscie:
Regarding issue #338 https://github.com/ActivityWatch/activitywatch/issues/338

### DIFF
--- a/src/components/InputTimeInterval.vue
+++ b/src/components/InputTimeInterval.vue
@@ -38,11 +38,16 @@ import moment from 'moment';
 
 export default {
   name: "input-timeinterval",
+  props: {
+      duration: {
+          type: Number,
+          default: 60 * 60
+      }
+  },
   data: () => {
     return {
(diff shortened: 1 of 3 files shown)


ErikBjare:
Looks great! 

Just one thing: the TimePeriodInput component is intended to be used in other places as well, where the setting should not be applied. I believe the proper way to handle that is to make the initial value an optional component prop (basically an argument) that is passed from the timeline view.

If you could fix that, that'd be great. If you don't know how props and stuff work, just let me know and I'll fix it shortly 👌

riscie:
I'll have a look :+1: (I'm an angular dev myself, so bare with me if this does not turn out as planned ^.^)

riscie:
I've updated the pr with what I think is what you were suggesting. Let me know if this works for you.

ErikBjare:
Excellent, that is exactly what I had in mind! Merging.

PR #183, opened 2020-04-08
started working on porting new Activity view to Android

ErikBjare:


ErikBjare:
@johan-bjareholt There's still some work left to do, but first I'd like to discuss my approach to dealing with the overlap between Android and desktop stuff. It isn't perfect, but it works and I'm not sure what would be the best way to do it.

src/store/modules/activity.ts
```diff
@@ -256,11 +290,13 @@ const actions = {
       state.buckets.browser_buckets.length > 0;
     const active_available = state.buckets.afk_buckets.length > 0;
     const editor_available = state.buckets.editor_buckets.length > 0;
+    const android_available = state.buckets.window_buckets[0].includes('android');
```
johan-bjareholt, line 293: Oh, the android bucket just have a normal bucket type?
That certainly makes things confusing.
Let's just hope that no one has 'android' in their hostname lol.
Can't you have something like ".startsWith('aw-watcher-android')" or something instead?

johan-bjareholt: Also, I think you need to check that "window_buckets" has a length of 1 or more before doing that check to avoid errors on desktop.

johan-bjareholt: Wouldn't it be easier if you in the "buckets" vuex instead had a "state.buckets.android_buckets" instead of here?
You could then also exclude the android buckets from the usual window buckets.

src/queries.ts
```diff
@@ -9,18 +9,26 @@ function querystr_to_array(querystr: string): string[] {
...
-  bid_window: string;
-  bid_afk: string;
-  bid_browsers?: string[];
+interface BaseQueryParams {
   filter_afk: boolean;
   include_audible?: boolean;
   classes: Record<string, any>;
 }
 
+interface DesktopQueryParams extends BaseQueryParams {
```
johan-bjareholt, line 18: Instead of having one for desktop and one for android, would it be possible to just have one but force "showAfk" to always be false when running on android so the afk_bucket is ignored?

src/store/modules/activity.ts
ErikBjare: >Oh, the android bucket just have a normal bucket type?

Yeah, bad initial choice on my part. I expect us to change this eventually in some Great Migration (as with currentwindow, afkstatus).

>Let's just hope that no one has 'android' in their hostname lol.
>Can't you have something like ".startsWith('aw-watcher-android')" or something instead?

Fixed.

>Also, I think you need to check that "window_buckets" has a length of 1 or more before doing that check to avoid errors on desktop.

Fixed.

>Wouldn't it be easier if you in the "buckets" vuex instead had a "state.buckets.android_buckets" instead of here?
>You could then also exclude the android buckets from the usual window buckets.

Fixed.

src/queries.ts
ErikBjare: There would still be separate queries since the events have different keys, so I don't think that'd make sense.

ErikBjare:
@johan-bjareholt Also note the thing in https://github.com/ActivityWatch/aw-webui/pull/183/commits/035d4cc66c4a2560133aab3afa8cb6d5427a1d44, looks like you broke it before.

ErikBjare:
@johan-bjareholt I think this should be merged now. There's still stuff that could be nicer but it works and I won't have more time to work on this anytime soon.

src/components/SelectableVisualization.vue
```diff
@@ -109,24 +113,56 @@ export default {
...
-    },
-    top_domains: function() {
-      return this.$store.state.activity.browser.top_domains;
-    },
-    top_urls: function() {
-      return this.$store.state.activity.browser.top_urls;
-    },
-    top_categories: function() {
-      return this.$store.state.activity.category.top;
+    visualizations: function() {
```
johan-bjareholt, line 116: Nice!

src/components/SelectableVisualization.vue
```diff
@@ -37,29 +44,26 @@ div
...
 </template>
 
-<style scoped lang="scss">
+<style lang="scss">
 .vis-style-dropdown-btn {
   position: absolute;
-  bottom: 0;
-  right: 0.5em;
+  top: 0.8em;
+  right: 1em;
```
johan-bjareholt, line 62: To be pixel perfect in line this should also be 0.8em.
I'll fix.

src/queries.ts
johan-bjareholt: Ah, didn't know

src/store/modules/activity.ts
```diff
@@ -69,15 +67,21 @@ const _state = {
...
     browser_buckets: 'all',
     editor_buckets: 'all',
   },
+
   buckets: {
-    afk_buckets: [],
-    window_buckets: [],
-    editor_buckets: [],
-    browser_buckets: [],
+    afk: [],
```
johan-bjareholt, line 80: Nice!

src/main.js
```diff
@@ -23,6 +23,11 @@ import 'typeface-varela-round';
 // Loads all the filters
 import './util/filters.js';
 
+// Create an instance of AWClient as this.$aw
+import awclient from './util/awclient.js';
+console.log(awclient);
```
johan-bjareholt, line 28: I'll remove.

src/store/modules/activity.ts
```diff
@@ -233,12 +250,30 @@ const actions = {
...
-    const data = await this._vm.$aw.query(periods, queries.dailyActivityQuery(bucket_id_afk));
+    const data = await this._vm.$aw.query(
+      periods,
+      queries.dailyActivityQuery(state.buckets.afk[0])
+    );
     const active_history = _.zipObject(
       periods,
       _.map(data, pair => _.filter(pair, e => e.data.status == 'not-afk'))
     );
+    console.log(active_history);
```
johan-bjareholt, line 261: I'll remove.

Issue #180, opened 2020-04-02
Start of week


skuzzymiglet:
By default, the week starts on Sunday. There should be an option to change this. I might make a PR

johan-bjareholt:
Good call, I've never realized but that seems to be true.

I guess this would be the solution
https://stackoverflow.com/a/18875953

johan-bjareholt:
Fixed on master, will be fixed in the next release!

PR #185, opened 2020-04-18
fix: Make week start on Monday instead of Sunday

johan-bjareholt:
Solves #180

### DIFF
--- a/src/views/activity/daily/ActivityDaily.vue
+++ b/src/views/activity/daily/ActivityDaily.vue
@@ -181,6 +181,21 @@ export default {
     periodReadable: function() {
       return moment(this.timeperiod.start).format(this.dateformat);
     },
+    periodLengthMoment: function() {
+      if (this.periodLength === 'day') {
+        return 'day';
+      } else if (this.periodLength === 'week') {
+        /* This is necessary so the week starts on Monday instead of Sunday */
+        return 'isoWeek';
+      } else if (this.periodLength === 'month') {
+        return 'month';
+      } else if (this.periodLength === 'year') {
+        return 'year';
+      } else {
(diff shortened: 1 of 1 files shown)


ErikBjare:
Why such a convoluted fix? Looks bug prone to me... 

Couldn't you just have had a `timeperiod === 'week'? 'isoWeek' : timeperiod`? I don't think you should use `this.timeperiod` in setDate.

src/views/activity/daily/ActivityDaily.vue
```diff
@@ -181,6 +181,21 @@ export default {
     periodReadable: function() {
       return moment(this.timeperiod.start).format(this.dateformat);
     },
+    periodLengthMoment: function() {
```
ErikBjare, line 184: This function seems redundant. It basically just does `s/week/isoWeek`.

src/views/activity/daily/ActivityDaily.vue
```diff
@@ -209,7 +224,7 @@ export default {
     },
     setDate: function(date, periodLength) {
       const new_date = moment(date)
-        .startOf(periodLength)
+        .startOf(this.periodLengthMoment)
```
ErikBjare, line 227: Don't use `this.periodLengthMoment` as it relies on `this.periodLength`, which makes it not respect the function parameter.

johan-bjareholt:
@ErikBjare First attempt was to actually have "Day" with a capital letter as lower case letter looks odd, but that had other side effects so I dropped that. But as I never actually fixed that your solution is shorter, but it has the flaw of not reporting an error when an invalid periodLength is set (which is possible when someone changes the periodLength in the URL).

Also, I was about to remove periodLength completely from setDate as it's pointless to pass it as an argument, but obviously forgot to do that.

PR #186, opened 2020-04-18
See host even if only editor buckets exist

### LINKED ISSUE #184: Why do I need an AFK and window watcher to get an activity report
Hi, first of all, I very much like this project and I'm moving from WakaTime to ActivityWatch.

But I couldn't figure out how I can access the activity reports page without enabling the AFK an window watchers? I'm using aw-server-rust built from source, which does not include those watchers, and I'm not interested in installing them.

Is there a reason for such high coupling? I guess since they are included in the default install they are assumed to be running, but I don't think they should get such special treatment that the report flat out don't work.

johan-bjareholt:
Fixes #184

### DIFF
--- a/src/components/Header.vue
+++ b/src/components/Header.vue
@@ -103,7 +103,7 @@ export default {
     //console.log(types_by_host);
 
     _.each(types_by_host, (types, hostname) => {
-      if (types.afk && types.window) {
+      if (hostname != 'unknown') {
         this.activityViews.push({
           name: hostname,
           hostname: hostname,
--- a/src/store/modules/activity_daily.ts
+++ b/src/store/modules/activity_daily.ts
@@ -56,8 +56,12 @@ const _state = {
     browser_buckets: 'all',
     editor_buckets: 'all',
   },
-  editor_buckets_available: [],
-  browser_buckets_available: [],
+  buckets: {
+    afk_buckets: [],
+    window_buckets: [],
+    editor_buckets: [],
+    browser_buckets: [],
+  },
 };
 
 function timeperiodsAroundTimeperiod(timeperiod: TimePeriod): TimePeriod[] {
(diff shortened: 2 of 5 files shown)


src/store/modules/buckets.js
```diff
@@ -3,13 +3,26 @@ const state = {
...
+function get_buckets_by_type(buckets, type) {
+  return _.map(
+    _.filter(buckets, bucket => bucket['type'] === type),
+    bucket => bucket['id']
+  );
+}
+
 // getters
 const getters = {
+  afkBuckets(state) {
```
ErikBjare, line 15: Nice!

src/store/modules/activity_daily.ts
```diff
@@ -25,6 +25,20 @@ function timeperiodToStr(tp: TimePeriod): string {
   return [start, end].join('/');
 }
 
+function get_buckets_by_type(buckets, type) {
```
ErikBjare, line 28: Nice, but duplicated code (also in `modules/buckets.js`). 

Move all of it there?

johan-bjareholt: That's the plan, hence the comment below

// TODO: re-use modules/buckets.js instead of doing a second getBuckets call

ErikBjare: Ah, could still have done an import for this particular helper function in the meantime, but whatever if you're going to fix it before merging.

src/views/activity/daily/ActivityDaily.vue
```diff
@@ -21,7 +21,7 @@ div
             |  Next
           icon(name="arrow-right")
     div.p-1
-      b-select(:value="periodLength", :options="['day', 'week', 'month']",
+      b-select(:value="periodLength", :options="['day', 'week', 'month', 'year']",
```
ErikBjare, line 24: Oooof, going to get some HTTP timeouts until aw-server-rust becomes the default.

src/components/Header.vue
```diff
@@ -103,7 +103,7 @@ export default {
     //console.log(types_by_host);
 
     _.each(types_by_host, (types, hostname) => {
-      if (types.afk && types.window) {
+      if (hostname != 'unknown') {
```
ErikBjare, line 106: Looks like this makes the `_.each` over types unnecessary?

src/views/activity/daily/ActivityDaily.vue
johan-bjareholt: Oops, commited by mistake.

src/components/Header.vue
johan-bjareholt: Not sure due to the android related code, don't want to break it by mistake.

ErikBjare: The Android view is getting merged with the normal view in https://github.com/ActivityWatch/aw-webui/pull/183 anyway, so feel free to remove that.

johan-bjareholt:
@ErikBjare Seems to work fine now and fixed some minor oddities, do you have any more comments or should I merge?

src/store/modules/activity_daily.ts
```diff
@@ -93,102 +97,128 @@ const getters = {
...
-      filterAFK,
-      classes,
-      filterCategories
-    );
-    const data = await this._vm.$aw.query(periods, q);
-    console.info(`Completed window query in ${moment().diff(start)}ms`);
-    commit('query_window_completed', data[0]);
+  async query_window({ state, commit }, { host, timeperiod, filterAFK, filterCategories }: QueryOptions) {
+    if (state.buckets.afk_buckets.length > 0 &&
+        state.buckets.window_buckets.length > 0) {
```
ErikBjare, line 125: Not really a fan of this check being here. Bucket-requirements should imo be checked in `ensure_loaded` and the `start_loading` thing should be called in the individual queries instead of one-fits-all.

Not a blocker for this PR, but would be nice if you could add that as a TODO comment in the code.

johan-bjareholt: > Not really a fan of this check being here. Bucket-requirements should imo be checked in ensure_loaded

Done

>  the start_loading thing should be called in the individual queries instead of one-fits-all.

Not sure what you mean by this?

src/components/Header.vue
johan-bjareholt: I think it'll be easier if you fix it in that change instead.

src/store/modules/activity_daily.ts
ErikBjare: Meant exactly what you changed it to, nice!

PR #190, opened 2020-04-25
Selectable activity views

### LINKED ISSUE #79: Customize activity summary
https://github.com/ActivityWatch/aw-webui/pull/77#issuecomment-398299381

johan-bjareholt:
Depends on #189 
Fixes #79 

This makes it possible to select any of the current visualizations in the "Summary" view so it becomes more of a "Favorites" view. You simply press the gear icon in the bottom right and select the type of visualization you want.

TODO:
- [x] Save selection in localstorage
- [x] Make the sunburst view not mess up alignment
- [x] Make views not selectable if they are not available (browser, editor, etc)
- [ ] Split up visualization components

Screenshot_2020-04-25 ActivityWatch


johan-bjareholt:
> I have some ideas for how to improve this further (I think the cog-dropdowns look a bit messy when everywhere), but I'll do that myself in another PR.

Yeah, they're not perfect but they're a good start and easy to implement.
I couldn't come up with a better way right now, but I'm sure there are better solutions.
It's also currently just a simple dropdown list, in the future it would be cool if we could have more advanced customization, maybe getting a modal like the event editor where you can customize things a bit more.

This solution in the grand scheme of things is definitely not perfect, but it's a good starting point to make it possible for users to have a custom dashboard.

PR #193, opened 2020-04-27
bugfix: Inconsistency when changing period length

johan-bjareholt:
Fixes https://github.com/ActivityWatch/activitywatch/issues/407

### DIFF
--- a/src/views/activity/Activity.vue
+++ b/src/views/activity/Activity.vue
@@ -25,7 +25,7 @@ div
                @change="(periodLength) => setDate(date, periodLength)")
     div.p-1(v-if="periodLength === 'day'")
       input.form-control(id="date" type="date" :value="date" :max="today"
-                         @change="setDate($event.target.value)")
+                         @change="setDate($event.target.value, periodLength)")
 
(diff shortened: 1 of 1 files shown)


src/views/activity/Activity.vue
```diff
@@ -25,7 +25,7 @@ div
                @change="(periodLength) => setDate(date, periodLength)")
     div.p-1(v-if="periodLength === 'day'")
       input.form-control(id="date" type="date" :value="date" :max="today"
-                         @change="setDate($event.target.value)")
+                         @change="setDate($event.target.value, periodLength)")
```
johan-bjareholt, line 28: This was the real dealbreaker, we were not passing the periodLength to this function here.

ErikBjare:
I made the same fix in #183 before you opened this PR.

johan-bjareholt:
> I made the same fix in #183 before you opened this PR.

@ErikBjare That's not enough, in the setDate function you are still using this.periodLengthMoment which is still the previos periodLength.

I'm merging this, the conflict should be easy to solve (or maybe git can even resolve it automatically).

PR #203, opened 2020-05-31
fix: Merge window and browser queries

johan-bjareholt:
Improves performance as events only needs to be fetched, flooded and afk-filtered once

Hopefully this doesn't break anything aw-android related, I don't think it should.

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -39,42 +39,6 @@ function canonicalEvents(params: DesktopQueryParams): string {
 
 const default_limit = 100; // Hardcoded limit per group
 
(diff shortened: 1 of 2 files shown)


ErikBjare:
Looks like it'll break if there aren't any browser events?

johan-bjareholt:
> Looks like it'll break if there aren't any browser events?

@ErikBjare It shouldn't in theory because then the browser_events array will be empty, but will extra check.

johan-bjareholt:
@ErikBjare Seems to work fine for me after deleting both my chrome and firefox buckets

PR #207, opened 2020-06-10
fix: fixed bug where opening Activity view didn't correctly use today as default date

ErikBjare:
Haven't actually tested it, but should work.

### DIFF
--- a/src/views/activity/Activity.vue
+++ b/src/views/activity/Activity.vue
@@ -17,14 +17,14 @@ div
                    variant="outline-dark")
             icon(name="arrow-left")
         b-select.px-2(:value="periodLength", :options="['day', 'week', 'month']",
-                 @change="(periodLength) => setDate(date, periodLength)")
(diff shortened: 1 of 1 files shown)


src/views/activity/Activity.vue
```diff
@@ -116,7 +116,11 @@ export default {
     host: String,
     date: {
       type: String,
-      default: get_today(),
+      // NOTE: This does not work as you'd might expect since the default is set on
+      // initialization, which would lead to the same date always being returned,
+      // even if the day has changed.
+      // Instead, use the computed _date.
+      //default: get_today(),
```
johan-bjareholt, line 123: No point in keeping the old code, could just as well be removed.

johan-bjareholt:
Test

src/views/activity/Activity.vue
ErikBjare: I'd like to keep it as documentation to explain why we aren't using the normal default and why you shouldn't use the date prop directly.

PR #210, opened 2020-06-27
feat: Better user experience when importing buckets

johan-bjareholt:
Instead of being redirected to an empty page the error is now shown on
the same page and the buckets list reloaded

### DIFF
--- a/src/views/Buckets.vue
+++ b/src/views/Buckets.vue
@@ -55,9 +55,13 @@ div
 
   b-card-group.deck
     b-card(header="Import buckets")
-      form(method="post", :action="$aw.baseURL + '/api/0/import'", enctype="multipart/form-data")
-        input(type="file", name="buckets.json")
-        input(type="submit", value="Import")
+      b-alert(v-if="import_error" show variant="danger" dismissable)
+        | {{ import_error }}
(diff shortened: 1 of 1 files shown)


johan-bjareholt:
@ErikBjare Seems to be some error with the screenshot test?

ErikBjare:
@johan-bjareholt Looks like it works as intended and found an error while trying to load the page. 

Did you look for the error in your own browser or try to run the test locally?

johan-bjareholt:
@ErikBjare Seems to break when you add the "dom" lib to tsconfig in aw-client-js.
It passes anyway if you just pass --skip-js-errors to testcase, but it's better to just fix the root cause.

ErikBjare:
Ah, as I mentioned on my late aw-client-js review then. 

And yeah, passing `--skip-js-errors` would kinda ruin the point of the test :)

johan-bjareholt:
> And yeah, passing --skip-js-errors would kinda ruin the point of the test :)

Well, it still runs and passes the test. It just ignores runtime errors in the browser.

johan-bjareholt:
Reverted the aw-client-js change and moved it into aw-webui which will always have a DOM, so it makes more sense to have it here.

PR #211, opened 2020-06-30
feat: started working on calendar visualization

ErikBjare:


ErikBjare:
This was kinda easy (although very rough around the edges to the point of almost being unusable). 

My main issue with FullCalendar is that there doesn't seem to be any easy way to zoom in/out during the day, so short events will practically disappear.

ErikBjare:
How it looks now:

image

ErikBjare:
I'll go ahead and merge this, as it's ready enough to be included in all dev builds.

PR #213, opened 2020-06-30
fix: cleaned up queries, unifying more into canonicalQuery

ErikBjare:


src/queries.ts
```diff
@@ -26,44 +27,63 @@ interface AndroidQueryParams extends BaseQueryParams {
...
+  const cat_filter_str = JSON.stringify(params.filter_classes);
+  const bid_window = isDesktopParams(params) ? params.bid_window : params.bid_android;
+
   return `
-    events  = flood(query_bucket("${params.bid_window}"));
-    not_afk = flood(query_bucket("${params.bid_afk}"));
-    not_afk = filter_keyvals(not_afk, "status", ["not-afk"]);
-    ${params.filter_afk ? 'events  = filter_period_intersect(events, not_afk);' : ''}
+    events = flood(query_bucket("${bid_window}"));
+    ${
```
johan-bjareholt, line 52: Not the most easily readable thing in the world, but still an improvement.

src/queries.ts
```diff
@@ -26,44 +27,63 @@ interface AndroidQueryParams extends BaseQueryParams {
...
+      isDesktopParams(params)
+        ? `not_afk = flood(query_bucket("${params.bid_afk}"));
+           not_afk = filter_keyvals(not_afk, "status", ["not-afk"]);`
+        : ''
+    }
+    ${isAndroidParams(params) ? 'events = merge_events_by_keys(events, ["app"]);' : ''}
```
johan-bjareholt, line 58: What's going on here, only merge them for android?
How does that make any sense, the events will be completely different on android and desktop?
If you do that here the I wouldn't really call this function "canonical" as the output is completely different.
Seems like it is merged anyway in appQuery, but differently for title and appname? (which looks correct, while this does not).

src/queries.ts
```diff
@@ -26,44 +27,63 @@ interface AndroidQueryParams extends BaseQueryParams {
...
+
+    ${
+      isDesktopParams(params) && params.filter_afk
+        ? 'events = filter_period_intersect(events, not_afk);'
+        : ''
+    }
+    ${params.classes ? `events = categorize(events, ${classes_str});` : ''}
```
johan-bjareholt, line 65: If we in the future also want to classify browser events we need to move this again, maybe it doesn't make sense to have it in canonicalEvents?

ErikBjare: The idea is to move browser events into canonical query, so no.

src/queries.ts
ErikBjare: They are merged here to avoid potentially returning several thousand events per day, same as happened in the android query before.

>If you do that here the I wouldn't really call this function "canonical" as the output is completely different.

If you want to call it something else then I'm open to suggestions.

>Seems like it is merged anyway in appQuery, but differently for title and appname? (which looks correct, while this does not).

It worked the same way before, I just moved the line.

PR #218, opened 2020-07-14
refactor: Split Settings into multiple views

johan-bjareholt:
Will make it easier to add more settings later, making each section more isolated from each other and therefore easier to grasp what's relevant and what's not.

### DIFF
--- a/src/components/CategoryEditTree.vue
+++ b/src/components/CategoryEditTree.vue
@@ -1,23 +1,23 @@
 <template lang="pug">
 div
-  div.row.py-2.cls
+  div.row.py-2.class
     div.col-8.col-md-4
       span(:style="{ marginLeft: (2 * depth) + 'em'}")
-        //| {{ cls.name.join(" ➤ ")}}
-      | #[span(v-if="depth > 0") ⮡] {{ cls.name.slice(depth).join(" ➤ ")}}
+        //| {{ _class.name.join(" ➤ ")}}
+      | #[span(v-if="depth > 0") ⮡] {{ _class.name.slice(depth).join(" ➤ ")}}
     div.col-4.col-md-8
       //span.d-none.d-sm-inline.d-md-none(:style="{ marginLeft: (2 * depth) + 'em'}")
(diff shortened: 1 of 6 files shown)


src/components/CategoryEditTree.vue
```diff
@@ -57,7 +57,7 @@ import 'vue-awesome/icons/edit';
 export default {
   name: 'CategoryEditTree',
   props: {
-    cls: Object,
+    _class: Object,
```
ErikBjare, line 60: Why the `cls` -> `_class` rename? 

`cls` is commonly used in Python (due to it being a reserved keyword). I'd personally prefer to not have underscores in names (unless they are private/local/special), but that's just me.

src/route.js
```diff
@@ -10,12 +10,13 @@ const ActivityWindow = () => import('./views/activity/ActivityWindow.vue');
 const ActivityBrowser = () => import('./views/activity/ActivityBrowser.vue');
 const ActivityEditor = () => import('./views/activity/ActivityEditor.vue');
 
+const Settings = () => import('./views/settings/Settings.vue');
+
```
ErikBjare, line 14: Why move the line? Will there be more routes?

src/components/CategoryEditTree.vue
johan-bjareholt: Because I googled for what cls might mean but couldn't find what it referred to (clear screen in DOS, common language specification, etc. were the results).
I understood it after a while, but it took me way too long.
I've seen klass and _class before so I went with _class.

src/route.js
johan-bjareholt: Oh sorry. First I thought so but then I decided not to. I'll revert it.

src/components/CategoryEditTree.vue
ErikBjare: Fair enough, I guess it's not that clear except in the Python-context:

```py
class A:
    @classmethod
    def f(cls, *args):
        return cls(*args)
```

PR #219, opened 2020-07-15
fix: fix timeline updating before new range is selected (#449 from ac…

billangli:
Hello! This is my first pull request, please feel free to let me know if you have feedback. I found this issue on the activitywatch repo (https://github.com/ActivityWatch/activitywatch/issues/449) and I thought that the pull request should go here.

### Before My Changes
Timeline View
Screenshot from 2020-07-15 13-41-58
Bucket View
Screenshot from 2020-07-15 13-42-22

### My Changes
I added a "Show Timeline" button that will send a request to the server if pressed. The default timeline range will still be showed when the page loads in the beginning. I also made the select boxes and the button horizontal because I thought that was better use of space.
Screenshot from 2020-07-15 14-30-16
In Range mode, the "Show Timeline" button will be disabled (grey) until the user has selected both dates.
Screenshot from 2020-07-15 14-30-40
Red text telling the user that the date range is invalid will also show if the first date is less than or equal to the second, the button is also greyed out.
Screenshot from 2020-07-15 14-30-58
Screenshot from 2020-07-15 14-31-24
The following two images are the new Bucket view.
Screenshot from 2020-07-15 14-57-30
Screenshot from 2020-07-15 14-58-00

### DIFF
--- a/src/components/InputTimeInterval.vue
+++ b/src/components/InputTimeInterval.vue
@@ -1,36 +1,44 @@
 <template lang="pug">
 div
   div
-    table
-      tr
-        th.pr-2
-          label(for="mode") Interval mode:
-        td
-          select(id="mode" v-model="mode")
-            option(value='last_duration') Last duration
-            option(value='range') Date range
-      tr(v-if="mode == 'last_duration'")
-        th.pr-2
-          label(for="duration") Show last:
-        td
-          select(id="duration" :value="duration", @change="valueChanged")
-            option(:value="15*60") 15min
-            option(:value="30*60") 30min
-            option(:value="60*60") 1h
-            option(:value="2*60*60") 2h
-            option(:value="4*60*60") 4h
-            option(:value="6*60*60") 6h
-            option(:value="12*60*60") 12h
-            option(:value="24*60*60") 24h
-      tr(v-if="mode == 'range'")
-        th.pr-2 Range:
-        td
-          input(type="date", v-model="start", @input="valueChanged")
-          input(type="date", v-model="end", @input="valueChanged")
+    div(class="settings-bar")
+      div
+        label(for="mode") Interval mode:
+        select(id="mode" v-model="mode")
+          option(value='last_duration') Last duration
+          option(value='range') Date range
+      div(v-if="mode == 'last_duration'")
+        label(for="duration") Show last:
+        select(id="duration" :value="duration" @change="changeDuration")
+          option(:value="15*60") 15min
+          option(:value="30*60") 30min
+          option(:value="60*60") 1h
+          option(:value="2*60*60") 2h
+          option(:value="4*60*60") 4h
+          option(:value="6*60*60") 6h
+          option(:value="12*60*60") 12h
+          option(:value="24*60*60") 24h
+      div(v-if="mode == 'range'")
+        | Range:
+        input(type="date", v-model="start")
+        input(type="date", v-model="end")
+      div
+        button.btn.btn-outline-dark(
+          type="button", 
+          :disabled="invalidDaterange || emptyDaterange",
+          @click="showTimeline"
+        ) Show Timeline
+    div(:style="{ color: 'red', visibility: invalidDaterange ? 'visible' : 'hidden' }")
+      | The selected date range is invalid.
 
 </template>
 
-<style scoped lang="scss"></style>
+<style scoped lang="scss">
+.settings-bar {
+  display: flex;
+  justify-content: space-between;
+}
+</style>
 
 <script>
 import moment from 'moment';
@@ -38,43 +46,46 @@ import moment from 'moment';
 export default {
   name: 'input-timeinterval',
   props: {
-    duration: {
+    defaultDuration: {
       type: Number,
       default: 60 * 60,
     },
   },
   data: () => {
     return {
-      now: moment(),
       mode: 'last_duration',
+      duration: null,
       start: null,
       end: null,
     };
+  },  
+  mounted: function() {
+    this.duration = this.defaultDuration;
   },
   computed: {
-    value: {
+    daterange: {
       get() {
         if (this.mode == 'range' && this.start && this.end) {
           return [moment(this.start), moment(this.end)];
         } else {
-          return [moment(this.now).subtract(this.duration, 'seconds'), moment(this.now)];
-        }
-      },
-      set(newValue) {
-        if (!isNaN(newValue)) {
-          // Set new now and duration
-          this.now = moment();
-          this.duration = newValue;
-        } else {
-          // Not required for mode='range', start and end set directly through v-model
+          return [moment().subtract(this.duration, 'seconds'), moment()];
         }
       },
     },
+    emptyDaterange() {
+      return this.mode == 'range' && !(this.start && this.end);
+    },
+    invalidDaterange() {
+      return this.mode == 'range' && moment(this.start) >= moment(this.end);
+    },
   },
   methods: {
-    valueChanged(e) {
-      this.value = e.target.value;
-      this.$emit('input', this.value);
+    showTimeline() {
+      this.$emit('update-timeline', this.daterange);
+    },
+    changeDuration(e) {
+      // This is only for last_duration, range mode changes daterange automatically
+      this.duration = e.target.value;
     },
   },
 };
--- a/src/components/InputTimeInterval.vue
+++ b/src/components/InputTimeInterval.vue
@@ -1,48 +1,50 @@
 <template lang="pug">
 div
   div
-    div(class="settings-bar")
-      div
-        label(for="mode") Interval mode:
-        select(id="mode" v-model="mode")
-          option(value='last_duration') Last duration
-          option(value='range') Date range
-      div(v-if="mode == 'last_duration'")
-        label(for="duration") Show last:
-        select(id="duration" :value="duration" @change="changeDuration")
-          option(:value="15*60") 15min
-          option(:value="30*60") 30min
-          option(:value="60*60") 1h
-          option(:value="2*60*60") 2h
-          option(:value="4*60*60") 4h
-          option(:value="6*60*60") 6h
-          option(:value="12*60*60") 12h
-          option(:value="24*60*60") 24h
-      div(v-if="mode == 'range'")
-        | Range:
-        input(type="date", v-model="start")
-        input(type="date", v-model="end")
-      div
-        button.btn.btn-outline-dark(
-          type="button", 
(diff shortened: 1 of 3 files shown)


src/components/InputTimeInterval.vue
```diff
@@ -1,80 +1,91 @@
...
+      return this.mode == 'range' && !(this.start && this.end);
+    },
+    invalidDaterange() {
+      return this.mode == 'range' && moment(this.start) >= moment(this.end);
+    },
   },
   methods: {
-    valueChanged(e) {
-      this.value = e.target.value;
-      this.$emit('input', this.value);
```
ErikBjare, line 77: Why remove the v-model behavior and replace with explicit events? Isn't it equivalent?

If I'm not mistaken, adding a simple check `if (!emptyDaterange() && !invalidDaterange()) {  this.$emit('input', this.value); }` would do it. Not sure why `this.value = e.target.value` was set.

src/components/InputTimeInterval.vue
```diff
@@ -1,80 +1,91 @@
...
       start: null,
       end: null,
     };
+  },  
+  mounted: function() {
+    this.duration = this.defaultDuration;
   },
   computed: {
-    value: {
+    daterange: {
```
ErikBjare, line 66: This rename (and removal of `set()`) breaks v-model.

src/components/InputTimeInterval.vue
```diff
@@ -1,80 +1,91 @@
...
   props: {
-    duration: {
+    defaultDuration: {
       type: Number,
       default: 60 * 60,
     },
   },
   data: () => {
     return {
-      now: moment(),
```
ErikBjare, line 48: Nice, this `now` was definitely not the right way to do it.

src/views/Timeline.vue
```diff
@@ -34,16 +34,13 @@ export default {
...
     },
   },
-  watch: {
-    daterange() {
-      this.getBuckets();
-    },
-  },
   mounted: function() {
-    this.getBuckets();
+    this.daterange = [moment().subtract(this.timeintervalDefaultDuration, "seconds"), moment()],
```
ErikBjare, line 38: Why manually construct and set daterange here? Doesn't `input-timeinterval` set `this.daterange` correctly? (I guess it might not, since you removed the `v-model` behavior)

src/components/InputTimeInterval.vue
```diff
@@ -1,80 +1,91 @@
...
+    invalidDaterange() {
+      return this.mode == 'range' && moment(this.start) >= moment(this.end);
+    },
   },
   methods: {
-    valueChanged(e) {
-      this.value = e.target.value;
-      this.$emit('input', this.value);
+    showTimeline() {
+      this.$emit('update-timeline', this.daterange);
```
ErikBjare, line 84: The function and event name here doesn't make sense. The component should not be coupled to the timeline.

src/components/InputTimeInterval.vue
```diff
@@ -1,80 +1,91 @@
...
-          label(for="mode") Interval mode:
-        td
-          select(id="mode" v-model="mode")
-            option(value='last_duration') Last duration
-            option(value='range') Date range
-      tr(v-if="mode == 'last_duration'")
-        th.pr-2
-          label(for="duration") Show last:
-        td
-          select(id="duration" :value="duration", @change="valueChanged")
```
ErikBjare, line 16: This select should probably have been a v-model to begin with.

src/components/InputTimeInterval.vue
```diff
@@ -1,80 +1,91 @@
...
+        | Range:
+        input(type="date", v-model="start")
+        input(type="date", v-model="end")
+      div
+        button.btn.btn-outline-dark(
+          type="button", 
+          :disabled="invalidDaterange || emptyDaterange",
+          @click="showTimeline"
+        ) Show Timeline
+    div(:style="{ color: 'red', visibility: invalidDaterange ? 'visible' : 'hidden' }")
```
ErikBjare, line 31: `div(v-show="invalidDaterange", style="color: 'red'")` is shorter and easier to read :)

src/components/InputTimeInterval.vue
```diff
@@ -1,80 +1,91 @@
...
+      div(v-if="mode == 'range'")
+        | Range:
+        input(type="date", v-model="start")
+        input(type="date", v-model="end")
+      div
+        button.btn.btn-outline-dark(
+          type="button", 
+          :disabled="invalidDaterange || emptyDaterange",
+          @click="showTimeline"
+        ) Show Timeline
```
ErikBjare, line 30: I really don't think this button is necessary, especially with my suggested changes to `valueChanged`.

src/components/InputTimeInterval.vue
```diff
@@ -1,80 +1,91 @@
...
     return {
-      now: moment(),
       mode: 'last_duration',
+      duration: null,
       start: null,
       end: null,
     };
+  },  
+  mounted: function() {
+    this.duration = this.defaultDuration;
```
ErikBjare, line 63: I think you can access props from the data constructor, so do that instead.

ErikBjare:
Oh, and it looks like GitHub Actions picked up some minor formatting issues. You can easily fix them with `make lint-fix`.

billangli:
Thank you for the detailed comments :+1:, it really helps especially because I'm new to Vue! Yeah I think it would be much easier if I kept the v-model and started from scratch.

src/components/InputTimeInterval.vue
billangli: I agree with not mentioning timeline in this component so that they are not coupled. I can make those changes. I included the button because in the issue, when looking at two date ranges (e.g. June 1 - June 2 and Apr. 1 - Apr. 2), after viewing the first date range, when you change June 1 to Apr. 1, aw-webui will render the timeline from Apr. 1 to Jun. 2, and the UI freezes for a couple of seconds while it is updating. Intuitively, I (and the issue author) expect the UI to wait for me to change both dates to Apr. 1 and Apr. 2 before updating the timeline. 

If I'm understanding you correctly, your suggested changes to valueChanged would still have the behaviour where the timeline updates before I finish inputting my second date, which is why I advocate for a button that sends an event up saying that the user has finished changing the date range. Let me know what you think!

src/components/InputTimeInterval.vue
billangli: I agree, I can change this.

src/components/InputTimeInterval.vue
billangli: I wanted to make the text take up space event when it is not invisible, but I can change this to keep it consistent with the rest of the UI

ErikBjare:
@billangli Starting from scratch sounds like overkill, just fix the things and then clean it up with rebase after :)

src/components/InputTimeInterval.vue
ErikBjare: Ah, right. I always mix up `visibility` with `display`. Feel free to keep it, and maybe add a comment why you're not using v-show :)

src/components/InputTimeInterval.vue
ErikBjare: Yeah, I get you. Not really sure what to do about that, maybe show the button only when the `mode` is `range`?

ErikBjare:
@billangli Oh, and if you're interested in getting paid for your contributions let me know. See https://github.com/ActivityWatch/activitywatch/issues/458

billangli:
Haha thanks for the offer! I don't think I've done enough yet so I'll pass on that. I just thought of this as a good opportunity to learn some full-stack development.

You probably saw my Bella repo, it was a silly hackathon project but I'd be happy to tell you about the idea. I also have a friend who is working on a research project that analyses speech patterns for diseases. It's signal processing so you might find it useful, let me see if he's available too.

src/components/InputTimeInterval.vue
billangli: Yeah now I think that having the button only for range would look good. I think adding the `max-duration` prop would be a good idea too.

billangli: I looked around to see if there's a configuration file to put `max-duration` in and it doesn't seem to be any except for Settings.vue, but I don't think users should be able to modify this value, should I just hard-code this value into InputTimeInterval.vue, Timeline.vue and Bucket.vue? 

By the way, I'm thinking of 31 days for `max-duration`. Thanks in advance for explaining all these little details to me!

billangli: @ErikBjare I forgot to tag you, I didn't know that it wouldn't send you a notification

ErikBjare: Sorry for the delayed reply. I'd just make it a prop in `InputTimeInterval` with a default of null, and then set it to ~30 days from `Timeline.vue` and `Bucket.vue`).

ErikBjare:
>I also have a friend who is working on a research project that analyses speech patterns for diseases

Small world. I've worked for [a startup](https://www.innovation.lu.se/en/lu-holding/portfolio-companies/voice-diagnostic) that does the same thing :slightly_smiling_face: 

Not my focus right now though so I'll pass on the intro for now, but thank you for the offer!

billangli:
@ErikBjare no worries! I totally get it, my friend mentioned to me that your areas are not as related as I thought haha.

billangli:
I made the changes that you suggested. I also changed the warnings to be more consistent with the rest of the UI. Let me know what you think.

Screenshot from 2020-07-24 22-59-18
Screenshot from 2020-07-24 22-59-38

src/views/Timeline.vue
```diff
@@ -40,7 +40,8 @@ export default {
     },
   },
   mounted: function () {
-    this.getBuckets();
+    (this.daterange = [moment().subtract(this.timeintervalDefaultDuration, 'seconds'), moment()]),
```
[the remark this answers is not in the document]
billangli: @ErikBjare I think I might have just forgot to check this and I wasn't being careful enough with the rebase, my bad. My intention was to look out for the case where `timeintervalDefaultDuration` is null because `localStorage.durationDefault` is null. I now removed this `mounted` function in Timeline.vue and made similar changes in Bucket.vue. The initial request will be called by `valueChanged` in a `mounted` function in InputTimelineInterval.vue.

nicolae-stroncea:
I just realized that Bucket Views (Raw Data->Open(any bucket)) has the same view and therefore the same problem. Are the same components used? If yes, then i think your PR will automatically solve the issue there too

billangli:
@nicolae-stroncea yep, Bucket view also uses the same component.

PR #217, opened 2020-07-13
feat: Add custom hovertext for editor summary and project

johan-bjareholt:

PR #222, opened 2020-07-16
Updated deps, fixed lints, applied prettier v2 style

ErikBjare:
We should probably merge these first (at the very least):

 - #218
 - #217 

Easier to rebase this on top of them (since this is mostly just a `make lint-fix`) than the other way around.


johan-bjareholt:
I fixed the merge conflicts with master and force pushed.

PR #224, opened 2020-07-29
fix: changed sometimes missing unicode character to icon in CategoryTree

ErikBjare:


johan-bjareholt:
@erikbjare Nice, this has been bugging me for a while actually

PR #229, opened 2020-10-05
feat: poll user satisfaction and calls to action

billangli:
# Description
The poll shows up after the user has used ActivityWatch for 30 days. If the user ignores/closes the poll, it will show up after 24 hours again for up to 3 times in total. The user can also choose to not show the poll again. Once the user submits a rating, the poll will not show up again. If the poll receives a rating that is >= 6, it will be a positive rating, otherwise negative. The respective followup alert shows up. 

Closes https://github.com/ActivityWatch/activitywatch/issues/291.

**Poll**
poll
**Positive Feedback Followup**
pos-feedback
**Negative Feedback Followup**
neg-feedback

# Questions
- I made this in a b-alert because it worked, @ErikBjare let me know what you feel about this
- I couldn't find the link to sign up for the newsletter
- I think the positive feedback followup might have too many suggested next-steps for the users, maybe we can take out some, I think 4-6 is a good number

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,13 +10,13 @@ module.exports = {
       testEnvironment: 'jsdom',
       transform: {
         '^.+\\.js$': 'babel-jest',
-        '^.+\\.vue$': 'vue-jest'
+        '^.+\\.vue$': 'vue-jest',
       },
       testMatch: ['**/test/**/*.test.js?(x)'],
       moduleNameMapper: {
         '^~/(.+)$': '<rootDir>/src/$1',
       },
-      modulePathIgnorePatterns: ['test/e2e/screenshot.test.js'] // Don't run this file in npm test
+      modulePathIgnorePatterns: ['test/e2e/screenshot.test.js'], // Don't run this file in npm test
     },
     {
       displayName: 'node',
@@ -25,7 +25,7 @@ module.exports = {
       testMatch: ['**/test/**/*.test.node.js?(x)'],
       moduleNameMapper: {
         '^~/(.+)$': '<rootDir>/src/$1',
-      }
-    }
-  ]
+      },
+    },
+  ],
 };
--- a/src/App.vue
+++ b/src/App.vue
@@ -4,6 +4,7 @@ div#wrapper
 
   div.container.aw-container.my-3.py-3
     error-boundary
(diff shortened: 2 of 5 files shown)


ErikBjare:
>I made this in a b-alert because it worked

That's fine, simpler logic than a modal :+1: 

>I think the positive feedback followup might have too many suggested next-steps for the users, maybe we can take out some, I think 4-6 is a good number

I agree. I'm not sure which actions are the most high-value (apart from the obvious like "donate" and "tell your friends"), but here's my thoughts on a few of them:

 - AlternativeTo: We have plenty of good reviews there already, not sure how much value it would add to have more of them.
 - Feedback form: I think we can comment out the feedback form for now. Getting more answers doesn't really help us at this time.
 - Newsletter: I think we'll take it out entirely. It's not really active. People who vote on the forum will have to register which then gets them notifications about announcements, so that's kinda equivalent and a better solution overall.

If you want to fix those things go ahead, or I can merge it straight away (looks good!) and decide on which actions to keep myself.

Excellent job! :tada:

ErikBjare (approved):
>I made this in a b-alert because it worked

That's fine, simpler logic than a modal :+1: 

>I think the positive feedback followup might have too many suggested next-steps for the users, maybe we can take out some, I think 4-6 is a good number

I agree. I'm not sure which actions are the most high-value (apart from the obvious like "donate" and "tell your friends"), but here's my thoughts on a few of them:

 - AlternativeTo: We have plenty of good reviews there already, not sure how much value it would add to have more of them.
 - Feedback form: I think we can comment out the feedback form for now. Getting more answers doesn't really help us at this time.
 - Newsletter: I think we'll take it out entirely. It's not really active. People who vote on the forum will have to register which then gets them notifications about announcements, so that's kinda equivalent and a better solution overall.

If you want to fix those things go ahead, or I can merge it straight away (looks good!) and decide on which actions to keep myself.

Excellent job! :tada:

billangli:
Thanks @ErikBjare! I think those are good points. I don't really have a strong preference on which actions to take out so I'll leave it up to you.

Let me know if there's anything else that needs to be worked on. I'm happy to help. I can probably get started again next week :smiley:

PR #230, opened 2020-10-07
docs: fixed comment in PeriodUsage.vue

alcsmadia:
Isn't is wrong? Thanks.


ErikBjare:
You're absolutely right. Thanks for the fix :+1:

PR #233, opened 2020-10-30
feat: started working on customizable views

ErikBjare:
Was surprisingly easy to get the basics working.

# TODO

 - [x] Add view
 - [x] Remove view
 - [x] Add visualization to view
 - [x] Remove visualization from view
 - [x] Migrate editor view
 - [x] Change visualization (to work as before)
 - [x] Save views
 - [x] Restore default views
 - [x] Set/rename view name
   - Done by adding a "new view" modal, created views still cannot be renamed
 - [x] Add notice for visualizations that are missing prerequisite data

# Extras (can wait for another PR)

 - [ ] Reorder visualizations in view
 - [ ] Add ability to set options for individual visualizations
     - Such as width/size, or visualization-specific options.

# Screenshot

image


ErikBjare:
Added the ability to remove visualizations, and an 'edit view' button:

image

image

ErikBjare:
@johan-bjareholt Ready for review :tada:

src/views/activity/ActivityWindow.vue
```diff
@@ -1,53 +0,0 @@
...
-  div.row.mb-6
-    div.col-md-6
-      h5 Top Applications
-      aw-summary(:fields="top_apps", :namefunc="e => e.data.app", :colorfunc="e => e.data.app", with_limit)
-    div.col-md-6
-      h5 Top Window Titles
-      aw-summary(:fields="top_titles", :namefunc="e => e.data.title", :colorfunc="e => e.data.app", with_limit)
-
-  div(v-if="periodLength == 'day'")
-    aw-sunburst-clock(:date="date", :afkBucketId="bucket_id_afk", :windowBucketId="bucket_id_window")
```
johan-bjareholt, line 12: There is now no way to use this visualization.
Personally I never use it, but it's worth mentioning.

src/views/activity/ActivityBrowser.vue
```diff
@@ -1,56 +0,0 @@
-<template lang="pug">
-div.mt-3
-  // TODO: Add back option to select a specific browser bucket
-  div(v-if="browserBuckets.length <= 0")
-    h6 No browser buckets available
-    small
-      | This feature requires a browser watcher.
-      | You can find a list of all watchers in #[a(href="https://activitywatch.readthedocs.io/en/latest/watchers.html") the documentation].
```
johan-bjareholt, line 8: It's a shame that this kind of notice is not there anymore.

src/views/activity/ActivityView.vue
```diff
@@ -0,0 +1,101 @@
...
+  },
+  methods: {
+    save() {
+      this.$store.dispatch('views/save');
+    },
+    discard() {
+      this.$store.dispatch('views/load');
+    },
+    remove() {
+      this.$store.commit('views/removeView', { view_id: this.view.id });
```
johan-bjareholt, line 74: Maybe this should gracefully fail if there is only a single view left and you try to delete it?

src/views/activity/Activity.vue
```diff
@@ -200,11 +209,19 @@ export default {
...
   mounted: async function () {
-    this.$store.dispatch('settings/load');
+    this.$store.dispatch('views/load');
+    this.$store.dispatch('categories/load');
     await this.refresh();
   },
 
   methods: {
+    addView: function () {
+      // TODO: Open modal to ask for options like id, and name
```
johan-bjareholt, line 219: I think that naming the view is a critical feature.
Seems very strange that the user has to remember what kind of visualizations is in an specific ID of a view.

johan-bjareholt (commented):
I get this error message on two occasions:
- When opening a custom view
- When looking at a custom view and pressing the "new view" button
```
Invalid prop: type check failed for prop "view_id". Expected String with value "5", got Number with value 5.
```

src/views/activity/ActivityBrowser.vue
ErikBjare: Yeah, I thought a bit about this since the same thing applies for the editor view.

Basically it would be nice if we'd have a way to indicate *why* a visualization cannot be selected, or alternatively, always let users select visualizations but then show a notice per-visualization if the requisite data is missing.

src/views/activity/ActivityWindow.vue
ErikBjare: I added a commented-out version to SelectableVisualization. We'll have to fix it later if there's interest.

src/views/activity/ActivityBrowser.vue
ErikBjare: Done!

src/views/activity/ActivityView.vue
ErikBjare: I think the current behavior is graceful enough: Removing the last view leads to zero views, and also no save button to persist the broken state. The 'New view' button still works.

src/util/transforms.js
```diff
@@ -0,0 +1,38 @@
+import _ from 'lodash';
+import moment from 'moment';
+
+// TODO: Move somewhere else, possibly turn into a serverside transform
```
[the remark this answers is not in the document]
ErikBjare: Wow. I did not have that functionality in mind when I wrote it, but you're right!

ErikBjare:
@johan-bjareholt I fixed the issues you mentioned in your review. I'll go ahead and merge this and will fix eventual issues in separate PRs.

PR #236, opened 2020-11-30
feat: added basic search feature

ErikBjare:
Adds a basic search feature.

# TODO

 - [x] Add to menu somewhere (where though?)
 - [ ] (extra) Manually set start and end
 - [ ] (extra) incremental search (don't search already queried timeperiods)
 - [ ] (extra) add barchart visualization for day/week
 - [ ] (extra) optimize query for long timeperiods

# Screenshots

image

image

### DIFF
--- a/src/views/Search.vue
+++ b/src/views/Search.vue
@@ -1,9 +1,11 @@
 <template lang="pug">
-
 div
   h3 Search
 
-  div.alert.alert-danger(v-if="error")
+  b-alert(style="warning" show)
+    | This feature is still in early development.
+
+  b-alert(v-if="error" show variant="danger")
     | {{error}}
 
   b-input-group(size="lg")
--- a/src/views/Search.vue
+++ b/src/views/Search.vue
@@ -1,0 +1,151 @@
+<template lang="pug">
+
+div
+  h3 Search
+
+  div.alert.alert-danger(v-if="error")
+    | {{error}}
+
+  b-input-group(size="lg")
+    b-input(v-model="pattern" placeholder="Regex pattern to search for")
+    b-input-group-append
+      b-button(type="button", @click="search()" variant="success")
+        icon(name="search")
+        | Search
+
+  div.d-flex.mt-1
+    span.mr-auto.small(style="color: #666") Hostname: {{hostname}}
+    b-button(size="sm", variant="outline-dark" style="border: 0" @click="show_options = !show_options")
+      span(v-if="!show_options")
+        | #[icon(name="angle-double-down")] Show options
+      span(v-else)
+        | #[icon(name="angle-double-up")] Hide options
+
+  div(v-if="show_options")
+    h4 Options
+    div Hostname
+      select(v-model="hostname")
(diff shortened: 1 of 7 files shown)


src/views/Search.vue
```diff
@@ -0,0 +1,151 @@
...
+    this.namefunc = this.summaryKeyFunc;
+    await this.$store.dispatch('buckets/ensureBuckets');
+    this.hostname = Object.keys(this.$store.getters['buckets/bucketsByHostname'])[0];
+  },
+  methods: {
+    search: async function () {
+      let query = canonicalEvents({
+        bid_window: 'aw-watcher-window_' + this.hostname,
+        bid_afk: 'aw-watcher-afk_' + this.hostname,
+        filter_afk: this.filter_afk,
```
johan-bjareholt, line 126: I don't think a "filter-afk" option is a good idea (assuming that we want the ability to delete events in the results).

Let's say someone searches for an event and wants to delete it, if that event spans both before and after and AFK intersection the user will delete the part both before and after that period which was possibly not intended.

So if we have a view somewhere where it's possible to delete event, that view should never be transformed. Adding more info to it such as with "split_url" or "categorize" is fine, but changing or removing keys/vals in the data or changing duration/timestamp on an event is not OK. And canonicalEvents does both flooding and filter_period_intersect so that's not good.

src/views/Search.vue
```diff
@@ -0,0 +1,151 @@
...
+  },
+  mounted: async function () {
+    this.colorfunc = this.summaryKeyFunc;
+    this.namefunc = this.summaryKeyFunc;
+    await this.$store.dispatch('buckets/ensureBuckets');
+    this.hostname = Object.keys(this.$store.getters['buckets/bucketsByHostname'])[0];
+  },
+  methods: {
+    search: async function () {
+      let query = canonicalEvents({
```
johan-bjareholt, line 123: Using canonicalEvents limits us to only delete events in the window bucket, I think it's better to drop the filter_afk support and just write a new query that simply fetches all events from a bucket and possibly categorizes them.

johan-bjareholt (changes requested):

src/views/Search.vue
ErikBjare: Yeah, I thought a bit about this but didn't know of a reasonable way to make it work with the upcoming merge of web-events (which was more or less point of refactoring into `canonicalEvents`).

I really don't want the user to have to think "which bucket could the thing I'm looking for be in?".

The simple solution here is to simply not let the user modify events in search results. If they want to delete/censor events there are much better ways than manually searching through them and deleting them one by one.

I'll make an example in aw-client that allows the user to censor/delete events.

src/views/Search.vue
ErikBjare: See https://github.com/ActivityWatch/aw-webui/pull/236#discussion_r537004344

ErikBjare:
Good enough to merge now. The ability to edit events in results was removed (for the very good reasons @johan-bjareholt mentioned).

PR #237, opened 2020-12-01
Allow horizontal scrolling (e.g. with touchpad) in timeline

NicoWeio:
I didn't get node-sass to run on my Manjaro machine, so I'll (ab)use your CI for this small change. :)

### DIFF
--- a/src/visualizations/VisTimeline.vue
+++ b/src/visualizations/VisTimeline.vue
@@ -51,6 +51,8 @@ export default {
           followMouse: true,
           overflowMethod: 'cap',
         },
+        horizontalScroll: true,
+        zoomKey: 'ctrlKey',
       },
     };
   },


ErikBjare:
Did you try 'npm rebuild node-sass' (or whatever other recommendation is spit out by node-sass)? If so, you'll probably have to use nvm or similar to get a lower node version. 

It's a pretty common issue: https://stackoverflow.com/questions/53125291/build-fails-npm-rebuild-node-sass-force

node-sass is actually deprecated, and I've wanted to make the switch to dart-sass, which I assume doesn't have these annoying build issues.

NicoWeio:
Wow, that indeed worked. No idea why I didn't find that solution while doing my googling. Thanks! 

Regarding horizontal scrolling, though: I managed to get it to work by setting `horizontalScroll: true`, but I could not find any configuration that would allow to both zoom (on scrolling vertically/normally) and scroll horizontally (= no UX change for those without mice/touchpads that allow for horizontal scrolling). I opened an [issue](https://github.com/visjs/vis-timeline/issues/817) for this, but maybe the configuration of my current prototype is better anyway: It scrolls horizontally for both input scroll directions, and allows zooming by using the `Ctrl` modifier. Vertical scrolling would also be possible.

ErikBjare:
Ah, that makes things a bit more difficult. I'm not sure the UX change is for the better, I don't really like the idea of requiring modifier keys to zoom, but I guess it's an option.

Nice that you opened an issue, I'll keep an eye on it.

>given that you're working on a complete overhaul of the timeline anyway.

Not really working, more like planning :slightly_smiling_face:

ErikBjare:
>Update: Oops, no build artifact. Gonna try something else…

Just a heads up, I added uploading of the build artifact in ba2374b :slightly_smiling_face:

PR #239, opened 2020-12-06
fix: fixed occurrences of `replace` that should be `replaceAll`

ErikBjare:
Fixes regex issue reported in https://forum.activitywatch.net/t/regex-still-doesnt-work/950/6

A real TIL moment for me personally, I had no idea `replace` only replaced the first occurrence... Did a quick `git grep 'replace'` and looks like these were the only ones.

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -9,6 +9,10 @@ function querystr_to_array(querystr: string): string[] {
     .map(l => l + ';');
 }
 
+function escape_doublequote(s: string) {
+  return s.replace(/"/g, '\\"');
+}
+
 interface BaseQueryParams {
   include_audible?: boolean;
   classes: Record<string, any>;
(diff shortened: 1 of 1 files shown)


ErikBjare:
Ugh, looks like `String.prototype.replaceAll` isn't in Node: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll

ErikBjare:
Hmm... LGTM warns "This does not escape backslash characters in the input" about the following:

```js
return s.replace(/"/g, '\\"');
```

I wonder what consequences this could have, if any?

ErikBjare:
@johan-bjareholt I'd love to have your input on the question in the previous comment.

But for now, this is an improvement, so merging.

PR #251, opened 2021-01-10
fix: fix category name change affecting sister category (ActivityWatch/activitywatch#361)

billangli:
The following changes address this issue: https://github.com/ActivityWatch/activitywatch/issues/361. 
* Use category ID to identify categories
* Delete category is moved from the edit modal to be its own button, it also has a confirmation modal now
* Delete category also deletes all descendents of that category (i.e. deletes its subtree)
* In the edit modal, it will no longer show categories in that category's subtree as its potential parent

Here are the new UI changes: 
Screenshot from 2021-01-09 20-06-00
Screenshot from 2021-01-09 20-06-14

The following code changes also allow two categories with the same path and name (e.g. Work -> Programming and Work -> Programming). I was originally thinking that I would have to change the backend to support this change. However, we don't have to change the backend if we disallow these duplicates. Then the web app can still query categories by using its full name (e.g. `['Work', 'Programming']`). Let me know if you think of any other solutions.

Edit: It is failing two tests because I removed the code from `build_category_hierarchy` that fills in missing parents. This is because before this change, each category's name (e.g. `['Work', 'Programming', 'ActivityWatch']`) already stores information about all of its parents, so you can keep only this information in storage and generate the parent categories Work and Work -> Programming. But now each category only keeps a reference to its immediate parent or children, which is why we can no longer generate parents. 

**Edit 2: I'm starting to think that an easier solution is to require all full names are unique and it would not need use ID, it would resolve the problems that I mentioned in the two paragraphs above.**

### DIFF
--- a/src/components/CategoryEditTree.vue
+++ b/src/components/CategoryEditTree.vue
@@ -24,7 +24,7 @@ div
       b-input-group.my-1(prepend="Name")
         b-form-input(v-model="editing.name")
       b-input-group(prepend="Parent")
-        b-select(v-model="editing.parent_id", :options="allOtherCategories")
+        b-select(v-model="editing.parentId", :options="allOtherCategories")
 
     hr
 
@@ -72,7 +72,7 @@ export default {
         id: null,
         name: null,
         rule: {},
-        parent_id: null,
+        parentId: null,
       },
     };
   },
@@ -84,14 +84,8 @@ export default {
         return { text: c.name.join('->'), value: c.id };
       });
 
-      const subtreeIds = [];
-      function findSubtree(node) {
-        subtreeIds.push(node.id);
-        if (node.children) {
-          node.children.forEach(c => findSubtree(c));
-        }
-      }
-      findSubtree(this._class);
+      const findSubtree = this.$store.getters['categories/find_subtree'];
+      const subtreeIds = findSubtree(this._class.id);
 
       return [{ value: -1, text: 'None' }].concat(
         entries.filter(c => !subtreeIds.includes(c.value))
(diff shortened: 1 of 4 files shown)


johan-bjareholt:
Two tests regarding classes seem to be failing, either there is some bug or the tests need to be fixed.

johan-bjareholt:
> Edit 2: I'm starting to think that an easier solution is to require all full names are unique and it would not need use ID, it would resolve the problems that I mentioned in the two paragraphs above.

Yes, I think this is better. The case you mentioned before with allowing two categories with the same name and parent ("Work -> Programming" and "Work -> Programming") doesn't seem like a real use-case. The important thing to fix is that for example two categories with the same name but different parent doesn't clash.

ErikBjare:
Sorry for the late review.

>I'm starting to think that an easier solution is to require all full names are unique and it would not need use ID

I'm leaning towards this too. 

I have a few questions about the diff, but those comments probably wouldn't be relevant if we did it this way anyway. (one of the concerns was how updating AW will affect existing instances where the `parentId`'s are missing)

I think simply changing the 'New category' button behavior to open the 'edit category' modal directly would resolve the issue (as described in https://github.com/ActivityWatch/activitywatch/issues/361#issuecomment-596180857).

---

And regarding moving the 'delete category' button, the previous UI of hiding it in the edit modal was very intentional (due to limited space on small screens, like on mobile), so I think it should be reverted.

billangli:
Thanks for the review, guys. Here is what I'll change:

- Show 'edit category' whenever a new category is created
- Make sure the full names (including ancestors) are unique
- Keep the old delete button (I do agree that it is crowded)

PR #252, opened 2021-01-20
Unify punctuation in satisfaction poll

NicoWeio:
I added full stops to every bullet point, as they all are full sentences, and I also fixed a conditional clause.


ErikBjare:
Nice, thanks!

PR #254, opened 2021-02-07
feat: started working on category colors

ErikBjare:
Lets users set colors on categories, and uses them for various visualizations.

Closes https://github.com/ActivityWatch/activitywatch/issues/416

# TODO

 - [ ] Make more visualizations use the new coloring logic
    - [x] Timeline (barchart)
    - [ ] Summary
      - [x] 'Top Apps' and 'Top Windows' (with the method described in https://github.com/ActivityWatch/activitywatch/issues/416#issuecomment-753461178)
      - [x] 'Top Categories' (runs the rule on the category name, when it should instead look up the category color)
      - [ ] Editor, browser, etc.
    - [x] Sunburst
    - [ ] Sunburst clock
 - [x] Rewrite `color.js` to TypeScript
 - [x] Add basic Vuex test example
 - [ ] Figure out how the upgrade process will work (where people have existing categories without `data.color` set)
 - [ ] Which colors should be set as default for the default categories?
 - [ ] Add a color picker/palett?
 - [ ] Fix all TODO comments

# Preview

Not a great deal of variety, but the basics are there (haven't done much else than work on ActivityWatch since I started the testing server...)

image


johan-bjareholt:
>  Figure out how the upgrade process will work (where people have existing categories without data.color set)

Upon page load, iterate over each category and simply set it if it's not already set?

ErikBjare:
I also wrote a basic test for the categories vuex module! (after some fiddling with the test config)

Should increase the code coverage somewhat, and make it easier to add more vuex tests.

ErikBjare:
There are a few more things before this is ready to ship, but I also want to ship #162 in the same release and I want to rebase on top of this, so merging.

Moving remaining tasks to an issue.

PR #255, opened 2021-02-09
feat: started working on auto refresh of activity page

ErikBjare:
Had this in my working tree for along time. Not ready and not sure if it's a good idea, but thought I should open a PR for it regardless.

### DIFF
--- a/src/views/activity/Activity.vue
+++ b/src/views/activity/Activity.vue
@@ -23,16 +23,25 @@ div
                    :disabled="nextPeriod() > today", variant="outline-dark")
             icon(name="arrow-right")
 
+
     div.mx-2(v-if="periodLength === 'day'")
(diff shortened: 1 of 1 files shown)


johan-bjareholt:
Personally I wouldn't care about an auto refresh every 30s, especially considering that weekly/monthly queries are very demanding.

Maybe if I have an activitywatch tab open which I go back to after an hour it would be cool if it reloaded, but I can also see how it could be annoying.

nicolae-stroncea:
A possibly less demanding option is to have a text field  "UI is outdated" that appears next to *Refresh* button if data has not been refreshed for past 5 minutes, thus letting user know they should *Refresh*. That said, I can see a use case for the refresh setting if somebody has the webpage in view/on another monitor and wants to see their activity live.

PR #246, opened 2021-01-02
test: started work on integration tests with cypress (including code coverage)

ErikBjare:
Apparently, TestCafe doesn't support code coverage. So trying out Cypress, which does.

Mostly just followed https://vuejsdevelopers.com/2020/07/20/code-coverage-vue-cypress/

## TODO

 - [ ] When running `npm run test:e2e`, process does not exit.
 - [ ] Doesn't really do any significant testing.
 - [ ] Doesn't run in CI.
 - [ ] Combine coverage reports.

PR #256, opened 2021-02-15
ci: added starting of aw-server(-rust) in CI for use in integration tests

ErikBjare:
Progress on true integration tests with fake data using https://github.com/ActivityWatch/aw-fakedata

The idea is to use this for true E2E tests and to make automated & accurate screenshots.

Related to #246

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -16,12 +16,19 @@ jobs:
     strategy:
       matrix:
         node-version: [12.x]
+        aw-version: ["v0.10.0"]
+        aw-server: ["aw-server", "aw-server-rust"]
 
     steps:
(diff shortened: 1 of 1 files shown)


ErikBjare:
@johan-bjareholt Anything I should think about when using aw-fakedata? Perhaps it needs some work to generate life-like data?

johan-bjareholt:
> Anything I should think about when using aw-fakedata?

Was a while since I wrote it, don't remember anything in particular.

> Perhaps it needs some work to generate life-like data? (similar to the demo button in the web UI)

Definitely, I just used it before to stress the web-ui weekly/monthly/yearly views. The data is unrealistic.

ErikBjare:
It works! 

Still work left to do to make screenshots look nice, but merging for now.

PR #258, opened 2021-02-22
feat: add productivity score for categories

ErikBjare:
Rebased and salvaged #162

Shouldn't require any changes to backend code.

## Tasks

 - [ ] Review what would make the best UX from previous discussions
 - [ ] How long you need to engage in a 1 point activity to get 1 point? (an hour?)
 - [x] Add computation of score for period
   - ```py
     totalScore = 0
     for event in events:
       totalScore += event.duration * get_score(event.category, allCategories)
     return totalScore / (60 * 60)  # from score-seconds to score-hours
     ```
 - [x] Add visualization (just a number of the total score today?)
 - [x] Make sure score is inherited correctly
 - [ ] What should scores be for the default categories?

### DIFF
--- a/src/components/CategoryEditModal.vue
+++ b/src/components/CategoryEditModal.vue
@@ -78,6 +78,8 @@ export default {
         parent: [],
         inherit_color: true,
         color: null,
+        inherit_score: true,
+        score: null,
       },
     };
   },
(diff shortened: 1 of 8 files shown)


johan-bjareholt:
>  How long you need to engage in a 1 point activity to get 1 point? (an hour?)

I never thought of it in that way at all.

Since you can do productive things outside of using your computer, to me it's more important to give a score of how productive you are on average rather than a sum of all your productive time.
So rather than to just say a number "Todays productivity score was 800", I had always imagined a scale from -1 to 1 which could in text say "Very productive" if it's for example above 0.7 or something.

nicolae-stroncea:
@johan-bjareholt  I think the current way is essentially identical to your way, except that the final value is unbounded.  The current formula essentially returns a **score-weighted average**, where the scores can be both negative and positive depending on what the scores the user sets, so user more productive -> total score increases,  user unproductive -> total score decreases.

If we want to bound the value, we can pass it through a `tanh(total_score)` function, which will map the values to [-1,1]. 

image

ErikBjare:
I added a really quick-n-dirty visualization in the latest commit:

image

ErikBjare:
Finally merging this. The remaining tasks will have to wait for another day.

PR #259, opened 2021-02-23
fixed undefined error

nicolae-stroncea:
This commit does 2 things:
* Does not access `old_class` attributes when it's undefined. This occurs because default categories do not have ids so old_class is undefined. This leads to a bunch of bugs and inconsistencies whenever a parent category is edited.
* Children are now renamed only when the parent category is actually renamed, instead of being renamed every time any properties of a category is modified.


nicolae-stroncea:
closing

PR #260, opened 2021-02-24
feat: Support import and export of categories

johan-bjareholt:
Todo:
- [x] Better design
- [x] Finish import

@ErikBjare Do you think that the categorization has a good format right now that we should proceed with, or are there any improvements you have in mind before essentially declaring it a standard format?

### DIFF
--- a/src/views/settings/CategorizationSettings.vue
+++ b/src/views/settings/CategorizationSettings.vue
@@ -6,6 +6,11 @@ div
     b-btn.ml-1(@click="restoreDefaultClasses", variant="outline-warning" size="sm")
       icon(name="undo")
       | Restore defaults
+    label.btn.btn-sm.ml-1.btn-outline-primary(style="margin: 0")
+      | Import
+      input(type="file" @change="importCategories" hidden)
+    b-btn.ml-1(@click="exportClasses", variant="outline-primary" size="sm")
+      | Export
   div
     | Rules for categorizing events. An event can only have one category. If several categories match, the deepest one will be chosen.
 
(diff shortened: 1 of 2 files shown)


src/views/settings/CategorizationSettings.vue
```diff
@@ -61,6 +66,33 @@ export default {
     restoreDefaultClasses: async function () {
       await this.$store.commit('categories/restoreDefaultClasses');
     },
+    exportClasses: function () {
+      console.log('Exporting categories...');
+
+      const text = JSON.stringify(this.classes_hierarchy);
```
ErikBjare, line 72: I don't think `classes_hierarchy` is the right thing to export. Just export the list of categories, as they are laid out in `localStorage`.

ErikBjare:
@johan-bjareholt I think the format is good, and just to be clear I expect the file/storage format to be:

```py
class Category:
    name: List[str]
    rule: Rule
    data: Dict | None
    id: int | None  # should probably be discarded/regenerated when exporting/importing

categories: List[Category]
```

johan-bjareholt:
@ErikBjare When exporting, would it make most sense to export what's currently being edited or what's saved?

ErikBjare:
@johan-bjareholt I thought a bit about that too, but I don't think we should support exporting unsaved changes. 

And on the flip side: we should maybe import everything as unsaved changes, to allow for review before saving?

johan-bjareholt:
@ErikBjare Seems to work well now! Ready for review

johan-bjareholt:
When exporting, it exports what's saved.
When importing, it imports as unsaved changes.

Format is { "categories": class dict }

ErikBjare (approved):

Issue #174, opened 2020-02-16
Improving the default categories


ErikBjare:
We've recently gotten PRs making improvements to the default categories. This is good, but writing good categorization rules can be tricky. 

Here are a few common pitfalls:

 - Apps that have non-unique names (i.e. Kate, Navigator, Files) 
   - Solution: Make sure that the whole string has to match
   - Solution: Only match certain event keys (not possible yet) 
 - Titles mentioning apps can get classified as apps. 
 - Loosely written expressions that over-match. It's a lot easier to find Uncategorized events than to find Miscategorized events. 

I'm sure there are more, these are only the ones off the top of my head. 

In this issue I will gladly accept suggestions for common apps and websites that should be added to the defaults, as well as suggested patterns for matching them in a way that minimizes the risk of misclassification.

nicolae-stroncea:
Suggestions for more apps to add to defaults:

*Email*: outlook, geary
*IM*: skype
*Movies*: netflix, primevideo, disneyplus
*Programming*: code(vscode), jupyter, readthedocs(not sure if always programming), terminal, tilix, guake

I think another potential pitfall is adding things to one category, that could belong to another category for somebody else. Ideally the default should be as general as possible.

ErikBjare:
Getting more nudges that people want better defaults (#261).

However, my current thinking is to leverage the new "export/import categories" feature to allow for community-curated categorization rules to be shared on the forum (which in turn will be linked to from the UI). 

This has a few benefits over simply changing the defaults (which wouldn't affect existing users), and encourages community experimentation and review over complex defaults.

iloveitaly:
> However, my current thinking is to leverage the new "export/import categories" feature to allow for community-curated categorization rules to be shared on the forum (which in turn will be linked to from the UI).

Oh, I love this! Maybe we could setup a separate repo to contain json category rules for various types?

That being said, I've always been a fan of smart defaults. Most folks don't want to fiddle with defaults and they want the first-run experience to generally be close-ish to what they are looking for. I think there's value in putting a little bit more thought and effort into improving the existing defaults, even if it causes some breakage for existing users (we could always freeze the existing defaults and link to them in the configuration UI if people people want to revert). Are you open to improving the existing defaults in addition to building a community-generated set of more complex defaults?

Mte90:
[...]

PR #261, opened 2021-02-24
More applications based on my must used app

Mte90:
Just few new apps

### DIFF
--- a/src/util/classes.ts
+++ b/src/util/classes.ts
@@ -28,7 +28,7 @@ export const defaultCategories: Category[] = [
   },
   {
     name: ['Work', 'Programming'],
-    rule: { type: 'regex', regex: 'GitHub|Stack Overflow|BitBucket|Gitlab|vim|Spyder|kate' },
+    rule: { type: 'regex', regex: 'GitHub|Stack Overflow|BitBucket|Gitlab|vim|Spyder|kate|Ghidra|Scite' },
   },
   {
     name: ['Work', 'Programming', 'ActivityWatch'],
@@ -40,14 +40,15 @@ export const defaultCategories: Category[] = [
     data: { color: '#0FF' },
   },
   { name: ['Media', 'Video'], rule: { type: 'regex', regex: 'YouTube|Plex|VLC' } },
+  { name: ['Media', 'Image'], rule: { type: 'regex', regex: 'Gimp|Inkscape' } },
   {
     name: ['Media', 'Social Media'],
     rule: { type: 'regex', regex: 'reddit|Facebook|Twitter|Instagram|devRant', ignore_case: true },
   },
-  { name: ['Media', 'Music'], rule: { type: 'regex', regex: 'Spotify|Deezer', ignore_case: true } },
+  { name: ['Media', 'Music'], rule: { type: 'regex', regex: 'Spotify|Deezer|Audacity', ignore_case: true } },
   {
     name: ['Comms', 'IM'],
-    rule: { type: 'regex', regex: 'Messenger|Telegram|Signal|WhatsApp|Rambox|Slack|Riot|Discord' },
+    rule: { type: 'regex', regex: 'Messenger|Telegram|Signal|WhatsApp|Rambox|Slack|Riot|Discord|Nheko' },
   },
   { name: ['Comms', 'Email'], rule: { type: 'regex', regex: 'Gmail|Thunderbird|mutt|alpine' } },
   { name: ['Uncategorized'], rule: { type: null }, data: { color: '#ccc' } },


src/util/classes.ts
```diff
@@ -40,14 +43,29 @@ export const defaultCategories: Category[] = [
     data: { color: '#0FF' },
   },
   { name: ['Media', 'Video'], rule: { type: 'regex', regex: 'YouTube|Plex|VLC' } },
+  { name: ['Media', 'Image'], rule: { type: 'regex', regex: 'Gimp|Inkscape' } },
```
iloveitaly, line 46: Wouldn't image editing be a work-related thing?

iloveitaly:
I hope we can get broader default categorizations like this merged in! I think this is super helpful.

src/util/classes.ts
Mte90: Well those tools are for editing so in this way is more specific and can be extended the video part also with video editors.
PS: I am italian and I like the nickname :-D

iloveitaly: That makes sense! @ErikBjare what do you think here? I have a couple changes along these lines I'd like to merge in.

My take on it is we should be more prescriptive about the initial categorization. Users can always customize it if they'd like to, but getting to a good set of defaults creates a better UX. In this case, I think most of the time if someone is editing images, they are doing it in a work-related way, so 'work > media' seems to make sense to me.

> PS: I am italian and I like the nickname :-D

Haha! Thanks :) I wish I was more authentically Italian, great to meet another italian dev :)

ErikBjare: @iloveitaly I made a comment in #174 with my thoughts, but I'm open to merging PRs like this if they cater to a broad audience :)

Although I do think the 'Media' category is for media *consumption* and not for media creation, that was my intent anyway :)

Mte90: Maybe a solution can be Work/Image Editing and Work/Video Editing and remove that Media/Image.
If it is fine for you I will updated the PR.

iloveitaly: I think that's a great idea!

johan-bjareholt:
I think this PR is pretty nicely structured and broad, just needs to fix a few linting errors.
After that I'm OK with merging this. @ErikBjare What do you think?

src/util/classes.ts
```diff
@@ -42,12 +47,26 @@ export const defaultCategories: Category[] = [
   { name: ['Media', 'Video'], rule: { type: 'regex', regex: 'YouTube|Plex|VLC' } },
   {
     name: ['Media', 'Social Media'],
-    rule: { type: 'regex', regex: 'reddit|Facebook|Twitter|Instagram|devRant', ignore_case: true },
+    rule: {
+      type: 'regex',
+      regex: 'reddit|Facebook|Twitter|Instagram|devRant',
+      ignore_case: true,
```
ErikBjare, line 53: An issue I had with this PR was that it had `ignore_case` set to true (which I think should be avoided in the defaults), but I now notice that was the case before anyway, so not really an issue for getting it merged.

src/util/classes.ts
```diff
@@ -42,12 +47,26 @@ export const defaultCategories: Category[] = [
...
+    },
+  },
+  {
+    name: ['Media', 'Music'],
+    rule: {
+      type: 'regex',
+      regex: 'Spotify|Deezer|Audacity',
```
ErikBjare, line 60: I feel like the same discussion regarding consumption vs production applies to Audacity. Perhaps move it?
```suggestion
      regex: 'Spotify|Deezer',
```

src/util/classes.ts
```diff
@@ -28,12 +28,17 @@ export const defaultCategories: Category[] = [
...
+      type: 'regex',
+      regex: 'GitHub|Stack Overflow|BitBucket|Gitlab|vim|Spyder|kate|Ghidra|Scite',
+    },
   },
   {
     name: ['Work', 'Programming', 'ActivityWatch'],
     rule: { type: 'regex', regex: 'ActivityWatch|aw-', ignore_case: true },
   },
+  { name: ['Work', 'Image Editing'], rule: { type: 'regex', regex: 'Gimp|Inkscape' } },
+  { name: ['Work', 'Video Editing'], rule: { type: 'regex', regex: 'Kdenlive' } },
```
ErikBjare, line 41:
```suggestion
  { name: ['Work', 'Image'], rule: { type: 'regex', regex: 'Gimp|Inkscape' } },
  { name: ['Work', 'Video'], rule: { type: 'regex', regex: 'Kdenlive' } },
  { name: ['Work', 'Audio'], rule: { type: 'regex', regex: 'Audacity' } },
  { name: ['Work', '3D'], rule: { type: 'regex', regex: 'Blender' } },
```

Removing the 'Editing' suffix to save space here. Not sure if this is the best call, but becomes repetitive otherwise (since the same suffix would be applied for Image/Video/Audio).

ErikBjare (commented):
Just a few things, will apply them myself and go ahead and merge.

ErikBjare:
Only screenshot thingy failing due to permissions. Merging.

Thanks for your contribution @Mte90! Sorry for not getting around to this sooner.

Mte90:
No problem :-)

PR #262, opened 2021-02-25
feat: progress on audible-as-active

ErikBjare:
I just banged this out without testing it at all, not sure if it'll work.

**Edit:** Huh, it wasn't too stupid after all. Seems to work...

Also re-adds the 'filter AFK' option.

Replaces #85 

Partially fixes https://github.com/ActivityWatch/activitywatch/issues/261

### Things to fix

 - [x] Prettier checkboxes at the bottom of the activity page
 - [x] More extensive testing?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,7 +19,7 @@
     "@typescript-eslint/explicit-function-return-type": ["off"],
     "@typescript-eslint/no-var-requires": ["off"],
     "@typescript-eslint/no-explicit-any": ["off"],
-    "@typescript-eslint/no-use-before-define": "warn",
+    "@typescript-eslint/no-use-before-define": ["error", {"functions": false, "classes": false}],
     "@typescript-eslint/no-unused-vars": ["warn", {"argsIgnorePattern": "^_"}],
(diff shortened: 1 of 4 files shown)


johan-bjareholt (approved):
Surprised that it was this clean to implement, can't really see any flaws. Looks good!

ErikBjare:
Did a bit more testing, seems to work fine. Merging when CI passes.

johan-bjareholt:
@ErikBjare Not working with aw-server-rust, there was one query method missing (forgotten which one, noticed a few days ago).

I think it should be pretty easy to fix though.

ErikBjare:
I assume it's period_union?

dtgay:
Seems like this feature doesn't work on Brave due to https://github.com/ActivityWatch/activitywatch/issues/461. Can anyone else confirm?

The tab activity _is_ being marked as audible, but I think since the browser watcher isn't working it fails to qualify for the AFK exception.

Looks like https://github.com/ActivityWatch/aw-webui/pull/375 might be a workaround but doesn't appear available in latest (0.12.1).

PR #277, opened 2021-03-30
fix: add google-chrome-stable to appnames

alialamine:
Add google-chrome-stable to the queries. This is the name of the chrome package when installed from the AUR


ErikBjare:
Nice, thanks!

PR #279, opened 2021-04-29
feat: added labels to sunburst, updated dependencies

ErikBjare:


ErikBjare:
I think I also accidentally updated the lockfile version to v2. 

Should be fine (https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json/), but explains the huge diff.

PR #281, opened 2021-05-03
Adding sourcemaps to vue dev builds

iloveitaly:
Super helpful for in-browser debugging.

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -17,6 +17,8 @@ module.exports = {
     });
   },
   configureWebpack: {
+    // sourcemaps are not enabled when `--watch` is used https://github.com/vuejs/vue-cli/issues/1806#issuecomment-832111894
+    devtool: 'source-map',
     resolve: {
       alias: {
         '~': path.resolve(__dirname, './src'),


ErikBjare:
I'm a bit surprised that sourcemaps aren't generated by default, but I do remember having issues with sourcemaps in the past. 

I found that there's a productionSourcemap option that is even set to true by default, so it seems like it should definitely be set in development builds: https://cli.vuejs.org/config/#productionsourcemap

I also skimmed https://github.com/vuejs/vue-cli/issues/1806 and there seems to be quite a few common issues both with setting this (relating to performance) and with leaving the defaults, not sure what setting this actually does.

Sourcemaps are indeed supposed to be enabled by default: https://github.com/vuejs/vue-cli/issues/1806#issuecomment-403924895

iloveitaly:
Interesting, if you run `vue-cli-service build` without any arguments sourcemaps are generated. However, if you run build with `--watch` sourcemaps are **not** generated. I [posted this](https://github.com/vuejs/vue-cli/issues/1806#issuecomment-832111894) in the issue thread to see if we get any interesting responses.

vue.config.js
```diff
@@ -17,6 +17,8 @@ module.exports = {
     });
   },
   configureWebpack: {
+    // sourcemaps are not enabled when `--watch` is used https://github.com/vuejs/vue-cli/issues/1806#issuecomment-832111894
```
ErikBjare, line 20: Excellent, was just going to suggest you add this comment :)

ErikBjare:
Merging!

PR #282, opened 2021-05-04
bug: run make install before running dev

iloveitaly:
The node packages need to be installed for dev to run properly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ install:
 uninstall:
 	rm -r node_modules/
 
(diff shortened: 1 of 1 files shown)


ErikBjare:
Indeed. Unfortunately `make` has no idea if the packages are already installed, and running `npm install` can take a while, so running it on every `make dev` is a bit excessive, which is why it's a seperate step :)

iloveitaly:
Makes sense! Thanks.

PR #283, opened 2021-05-04
refactor: separate brave browser events into their own category

iloveitaly:
It looks like this was partially done, so I just cleaned it up. Also added the name of brave browser on macOS.

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -131,9 +131,6 @@ const browser_appnames = {
     // Pre-releases
     'Google-chrome-beta',
     'Google-chrome-unstable',
-
-    // Brave (should this be merged with the brave entry?)
-    'Brave-browser',
   ],
   firefox: [
     'Firefox',
@@ -147,7 +144,7 @@ const browser_appnames = {
     'Nightly',
   ],
   opera: ['opera.exe', 'Opera'],
-  brave: ['brave.exe'],
+  brave: ['Brave-browser', 'Brave Browser', 'brave.exe'],
   edge: [
     'msedge.exe', // Windows
     'Microsoft Edge', // macOS


ErikBjare:
Cool! 

Short Q: what is the name of the browser bucket created by aw-watcher-web when you use Brave? I think we used to have some issues with identifying if it's actually running in Brave or if it's indistinguishable from Chrome (they've tried to avoid fingerprinting to a fault...)

iloveitaly:
> what is the name of the browser bucket created by aw-watcher-web when you use Brave

I'm not using the browser extensions right now, but here's what is being reported by the window watcher:

```
2021-05-05 11:46:25 [DEBUG]: {'app': 'Brave Browser', 'url': 'https://www.google.com/?gws_rd=ssl', 'title': 'Google', 'incognito': False}  (aw_watcher_window.main:71)
```

ErikBjare:
I think this might break things (although things were probably already broken), see: https://github.com/ActivityWatch/activitywatch/issues/461

iloveitaly:
@ErikBjare I didn't read the thread in detail, but I don't see how this would break anything: it should just properly categorize brave browser usage to a dedicated category instead of setting it to chrome.

Feel free to close this out if you feel this isn't useful!

ErikBjare:
Basically the "brave" field is currently unused in code since the web extension can't detect if its installed in Brave (it pretends to be Chrome). Therefore, using aw-watcher-web in Brave will report to the aw-watcher-web-chrome bucket instead of the aw-watcher-web-brave bucket (which is what the mapping is used for).

However, as @johan-bjareholt mentioned in https://github.com/ActivityWatch/activitywatch/issues/461 we might be able to address this by attributing Brave the window with the aw-watcher-web-chrome bucket (move all 'brave' entries to the 'chrome' key). 

It's possible that there are already Brave users on some platform (with the specific appname this PR moves) who've effectively already had this workaround, and for them this PR would break things.

However, this comes with problems if the watcher is installed in both browsers. Since both watchers would report to the same bucket, messing up the data.

Not sure what the right thing to do is here, maybe we should just merge this and fix Brave detection in aw-watcher-web. Will probably wait with merging this until there's a PR for the needed fixes to aw-watcher-web :)

iloveitaly:
Makes sense! Thanks for the explanation.

sanderlienaerts:
@ErikBjare Looked into this. The PR is ready for merge.

Used library `ua_parser` in `aw-watcher-web` includes Brave detection.

aw-watcher-web: https://github.com/ActivityWatch/aw-watcher-web/blob/0b289c4208f308050979fc729a69cf0c57dd7d8f/src/client.js#L4
ua-parser-js: https://github.com/faisalman/ua-parser-js/blob/b29a9a7ffbcb0fd0e99c5a49150c2ddb6fa91102/src/main/ua-parser.js#L1004)

PR #284, opened 2021-05-04
feat: improve formatting of query range in Activity view

iloveitaly:
This makes it much more explicit what range of data you are viewing. A week number isn't very helpful—I have no idea what range of days a number corresponds to. 

Here's what it looks like:

image

### DIFF
--- a/src/views/activity/Activity.vue
+++ b/src/views/activity/Activity.vue
@@ -1,6 +1,7 @@
 <template lang="pug">
 div
-  h3.mb-0 Activity for {{ timeperiod | friendlyperiod }}
+  h3 Activity
+    span.text-secondary.aw-period-range  {{ timeperiod | friendlyperiod }}
 
   div.mb-2
     ul.list-group.list-group-horizontal-md.mb-3(style="font-size: 0.9em; opacity: 0.7")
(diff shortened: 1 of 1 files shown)


ErikBjare:
The header will probably be too long on small displays (phones) if the entire range is included.

Perhaps you could instead keep the `W` formatting and show the range on the line below if the view is weekly/monthly?

iloveitaly:
What if we remove 'for' and use some styling to separate the title from the range?

image

image

johan-bjareholt:
Personally I actually look at the week numbers quite often, but that's a personal preference and I wouldn't be surprised if the majority don't use week numbers very often (as it's a pretty Swedish thing to do).

Another option would be to keep it as is, but add a "From" and "Until" section beneath as someone on the forum has requested:

image
https://forum.activitywatch.net/t/add-a-start-afk-and-end-time-for-a-day-to-the-dashboard/1220

iloveitaly:
@johan-bjareholt I'm good with either approach. At least in the US, I haven't seen a "week number" in any UI I can think of. Not so say we should optimize for a US-centric approach, I just don't have a good sense of what most people in the current userbase of activitywatch would find useful.

iloveitaly:
@johan-bjareholt @ErikBjare  Curious if you've had any updated thoughts here. The `W` notation continues to be annoying for me, would love to get a change here merged upstream.

johan-bjareholt:
I think I'm an minority that prefers week numbers, so I think that these changes are probably an improvement for most people. I'm fine with merging this once the merge conflict is fixed.

ErikBjare:
I'm working on merging this.

ErikBjare:
Rebased on master

iloveitaly:
Woohoo! Awesome.

PR #286, opened 2021-05-05
feat: more categorization and support for select_keys

iloveitaly:
Similar to [this PR](https://github.com/ActivityWatch/aw-webui/pull/261) (and as [discussed here](https://github.com/ActivityWatch/aw-webui/issues/174) this contains a additional app categorizations. 

Some additional changes:

* Added `select_keys` to the rule. This is helpful for matching on the domain name of an event.
* Added prettier colors (screenshot below)

image

### DIFF
--- a/src/util/classes.ts
+++ b/src/util/classes.ts
@@ -24,14 +24,14 @@ export interface Category {
   children?: Category[];
 }
 
-// https://colorhunt.co/palette/4802
-const COLOR_GREEN = '#96cd39',
+const COLOR_GREEN = '#0F0',
   COLOR_SUPER_GREEN = '#54e346',
   COLOR_BRIGHT_GREEN = '#A8FC00',
   COLOR_UNCAT = '#CCC',
+  COLOR_YELLOW = '#FCC400',
   COLOR_SLIGHTLY_YELLOW = '#f5ff65',
   COLOR_ORANGE = '#ffba47',
-  COLOR_RED = '#ff5b44';
+  COLOR_RED = '#F80';
 
 // The default categories
 // Should be run through createMissingParents before being used in most cases.
@@ -108,7 +108,7 @@ export const defaultCategories: Category[] = [
       regex: 'reddit|Facebook|Twitter|Instagram|devRant|LinkedIn',
       ignore_case: true,
     },
-    data: { color: COLOR_RED },
+    data: { color: COLOR_YELLOW },
   },
   {
     name: ['Media', 'Podcasts'],
--- a/src/util/classes.ts
+++ b/src/util/classes.ts
@@ -9,6 +9,7 @@ export interface Rule {
   type: 'regex' | 'none';
   regex?: string;
   ignore_case?: boolean;
+  select_keys?: string[];
 }
 
 export interface Category {
@@ -23,26 +24,68 @@ export interface Category {
   children?: Category[];
 }
 
-const COLOR_UNCAT = '#CCC';
+// https://colorhunt.co/palette/4802
+const COLOR_GREEN = '#96cd39',
+  COLOR_SUPER_GREEN = '#54e346',
+  COLOR_BRIGHT_GREEN = '#A8FC00',
+  COLOR_UNCAT = '#CCC',
+  COLOR_SLIGHTLY_YELLOW = '#f5ff65',
+  COLOR_ORANGE = '#ffba47',
+  COLOR_RED = '#ff5b44';
 
 // The default categories
 // Should be run through createMissingParents before being used in most cases.
 export const defaultCategories: Category[] = [
   {
-    name: ['Work'],
-    rule: { type: 'regex', regex: 'Google Docs|libreoffice|ReText' },
-    data: { color: '#0F0' },
+    name: ['Work', 'Writing'],
+    rule: { type: 'regex', regex: 'Google Docs|Google Sheets|libreoffice|ReText' },
+    data: { color: COLOR_GREEN },
+  },
+  {
+    name: ['Work', 'Writing'],
+    rule: {
+      type: 'regex',
+      select_keys: ['app'],
+      regex: 'LibreOffice|TextEdit|MacDown|Obsidian|TextEdit',
+    },
+    data: { color: COLOR_GREEN },
+  },
+  {
+    name: ['Work', 'General'],
+    rule: {
+      type: 'regex',
+      regex: 'Preview|Finder|Todoist|1Password|Soulver|System Preferences|VNC Viewer|Streaks',
+      select_keys: ['app'],
+    },
+    data: { color: COLOR_SLIGHTLY_YELLOW },
   },
   {
     name: ['Work', 'Programming'],
     rule: {
       type: 'regex',
-      regex: 'GitHub|Stack Overflow|BitBucket|Gitlab|vim|Spyder|kate|Ghidra|Scite',
+      regex:
+        'GitHub|Stack Overflow|BitBucket|Gitlab|vim|Spyder|kate|iTerm|Hyper|Alacritty|Script Editor|Ghidra|Scite|jetbrains-idea|jetbrains-pycharm',
     },
+    data: { color: COLOR_SUPER_GREEN },
+  },
+  {
+    name: ['Work', 'Programming'],
+    rule: { type: 'regex', select_keys: ['url'], regex: 'github.com' },
+    data: { color: COLOR_SUPER_GREEN },
+  },
+  {
+    name: ['Work', 'Programming'],
+    rule: {
+      type: 'regex',
+      select_keys: ['app'],
+      regex: 'Code|Sublime Text|TextMate|iTerm|Script Editor|Base|Postico|Sequel Ace',
+    },
+    data: { color: COLOR_SUPER_GREEN },
   },
   {
     name: ['Work', 'Programming', 'ActivityWatch'],
     rule: { type: 'regex', regex: 'ActivityWatch|aw-', ignore_case: true },
+    data: { color: COLOR_SUPER_GREEN },
   },
   { name: ['Work', 'Image'], rule: { type: 'regex', regex: 'Gimp|Inkscape' } },
   { name: ['Work', 'Video'], rule: { type: 'regex', regex: 'Kdenlive' } },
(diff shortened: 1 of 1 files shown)


iloveitaly:
@ErikBjare friendly reminder on this PR :)

Before I fix the last failing test, I want to make sure this is the sort of thing you'd be ok merging in.

src/util/classes.ts
```diff
@@ -20,26 +21,63 @@ export interface Category {
...
+      regex:
+        'GitHub|Stack Overflow|BitBucket|Gitlab|vim|Spyder|kate|iTerm|Hyper|Alacritty|Script Editor|Ghidra|Scite',
     },
+    data: { color: COLOR_SUPER_GREEN },
+  },
+  {
+    name: ['Work', 'Programming'],
+    rule: { type: 'regex', select_keys: ['url'], regex: 'github.com' },
+    data: { color: COLOR_SUPER_GREEN },
+  },
```
ErikBjare, line 67: Does having duplicate keys work correctly in the web UI? (editing etc)

src/util/classes.ts
```diff
@@ -48,27 +86,27 @@ export const defaultCategories: Category[] = [
   {
     name: ['Media', 'Games'],
     rule: { type: 'regex', regex: 'Minecraft|RimWorld' },
-    data: { color: '#F80' },
+    data: { color: COLOR_RED },
```
ErikBjare, line 89: Was orange, now red (intentional/improvement?)

src/util/classes.ts
```diff
@@ -48,27 +86,27 @@ export const defaultCategories: Category[] = [
...
   {
     name: ['Media', 'Social Media'],
     rule: {
       type: 'regex',
-      regex: 'reddit|Facebook|Twitter|Instagram|devRant',
+      regex: 'reddit|Facebook|Twitter|Instagram|devRant|LinkedIn',
       ignore_case: true,
     },
-    data: { color: '#FCC400' },
+    data: { color: COLOR_RED },
```
ErikBjare, line 103: Was yellow, now red.

src/util/classes.ts
```diff
@@ -48,27 +86,27 @@ export const defaultCategories: Category[] = [
...
   },
   {
     name: ['Media', 'Music'],
     rule: {
       type: 'regex',
-      regex: 'Spotify|Deezer',
+      regex: 'Spotify|Deezer|Amazon Music',
       ignore_case: true,
     },
     data: { color: '#A8FC00' },
```
ErikBjare, line 112: Can this color be lifted out into a variable too?

src/util/classes.ts
```diff
@@ -20,26 +21,63 @@ export interface Category {
   children?: Category[];
 }
 
-const COLOR_UNCAT = '#CCC';
+// https://colorhunt.co/palette/4802
+const COLOR_GREEN = '#96cd39',
+  COLOR_SUPER_GREEN = '#54e346',
+  COLOR_UNCAT = '#CCC',
+  COLOR_ORANGE_GREEN = '#f5ff65',
```
ErikBjare, line 28: Maybe rename/change this slightly to a yellow?

src/util/classes.ts
iloveitaly: It does seem to work from my testing!

src/util/classes.ts
iloveitaly: Fixed.

src/util/classes.ts
iloveitaly: Fixed.

src/util/classes.ts
iloveitaly: I can revert it back, but I think the dark red looks ok:

image

Maybe just personal preference on my end, so happy to revert if you'd like.

src/util/classes.ts
```diff
@@ -23,26 +24,68 @@ export interface Category {
...
-    data: { color: '#0F0' },
+    name: ['Work', 'Writing'],
+    rule: { type: 'regex', regex: 'Google Docs|Google Sheets|libreoffice|ReText' },
+    data: { color: COLOR_GREEN },
+  },
+  {
+    name: ['Work', 'Writing'],
+    rule: {
+      type: 'regex',
+      select_keys: ['app'],
```
ErikBjare, line 48: So this is set, but there is no way to change it in the UI? 

All `select_keys` should probably be removed from the default categories until the UI supports it.

src/util/classes.ts
```diff
@@ -23,26 +24,68 @@ export interface Category {
...
 
-const COLOR_UNCAT = '#CCC';
+// https://colorhunt.co/palette/4802
+const COLOR_GREEN = '#96cd39',
+  COLOR_SUPER_GREEN = '#54e346',
+  COLOR_BRIGHT_GREEN = '#A8FC00',
+  COLOR_UNCAT = '#CCC',
+  COLOR_SLIGHTLY_YELLOW = '#f5ff65',
+  COLOR_ORANGE = '#ffba47',
+  COLOR_RED = '#ff5b44';
```
ErikBjare, line 34: Please revert these to previous colors (but good to keep them as constants!)

ErikBjare (changes requested):
Just a couple minor things and I'll merge :)

src/util/classes.ts
```diff
@@ -23,26 +24,68 @@ export interface Category {
...
+    },
+    data: { color: COLOR_SLIGHTLY_YELLOW },
   },
   {
     name: ['Work', 'Programming'],
     rule: {
       type: 'regex',
-      regex: 'GitHub|Stack Overflow|BitBucket|Gitlab|vim|Spyder|kate|Ghidra|Scite',
+      regex:
+        'GitHub|Stack Overflow|BitBucket|Gitlab|vim|Spyder|kate|iTerm|Hyper|Alacritty|Script Editor|Ghidra|Scite',
```
soxofaan, line 67: please consider adding `jetbrains-pycharm` and `jetbrains-idea` to this list

iloveitaly: @soxofaan are these the application names on macos? Doesn't seem like it. Make a [suggestion](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) and I'll include it in this PR.

src/util/classes.ts
iloveitaly: Yes, that's right.

You can't set them via the UI, but you can import a config file which contains them. I think we should add `select_keys` to the default config since you can 'export, modify, import' and to mutate these keys _and_ it makes it more explicit how we are deciding which app is in use.

src/util/classes.ts
iloveitaly: Reverted!

iloveitaly:
@ErikBjare finally fixed this. Collapsed the commits & rebased. Hopefully this is what we need to get this in!

src/util/classes.ts
```diff
@@ -85,10 +137,29 @@ export const defaultCategories: Category[] = [
     name: ['Comms', 'IM'],
     rule: {
       type: 'regex',
-      regex: 'Messenger|Telegram|Signal|WhatsApp|Rambox|Slack|Riot|Element|Discord|Nheko',
+      regex:
+        'Messenger|Messages|Discord|Telegram|Signal|WhatsApp|Rambox|Slack|Riot|Element|Discord|Textual|Nheko|Texts',
+    },
+    data: { color: COLOR_ORANGE },
```
ErikBjare, line 143: Why override the `#9FF` comms color with something very different?

```suggestion
```

src/util/classes.ts
```diff
@@ -85,10 +137,29 @@ export const defaultCategories: Category[] = [
...
+  },
+  {
+    name: ['Comms', 'Email'],
+    rule: { type: 'regex', regex: 'Gmail|Thunderbird|mutt|alpine' },
+    data: { color: COLOR_ORANGE },
```
ErikBjare, line 148: Same as above.

```suggestion
```

src/util/classes.ts
```diff
@@ -85,10 +137,29 @@ export const defaultCategories: Category[] = [
...
+  },
+  {
+    name: ['Comms', 'Meetings'],
+    rule: { type: 'regex', regex: 'Zoom|Calendar|Cron' },
+    data: { color: COLOR_SLIGHTLY_YELLOW },
```
ErikBjare, line 153:
```suggestion
```

src/util/classes.ts
```diff
@@ -85,10 +137,29 @@ export const defaultCategories: Category[] = [
...
+  },
+  {
+    name: ['Web Browsing'],
+    rule: {
+      type: 'regex',
+      regex: 'Chrome|Safari|FireFox|Brave',
+      select_keys: ['app'],
```
ErikBjare, line 160:
```suggestion
```

src/util/classes.ts
```diff
@@ -197,6 +268,7 @@ function pickDeepest(categories: Category[]) {
   return _.maxBy(categories, c => c.name.length);
 }
 
+// TODO this is only used colorize the categories, all categorization is done on the backend
```
ErikBjare, line 271:
```suggestion
// NOTE: this is only used to colorize the categories, all actual categorization is done on the backend
```

src/util/classes.ts
ErikBjare: But that's very confusing for the user. They have no indication there is even a `select_keys` for those categories, right? (unless they export and inspect)

I really don't want to add this specifier to the default rules without:

 1. UI support for it
 2. More thorough testing of these new "multi-rule" categories in the UI.
     - What happens in the UI if a category with multiple rules has children?
     - What happens when you rename a category with multiple rules? Or assign it to a different parent?
     - What happens to `color` keys when you have multiple rules?
     - Ideally, a category with multiple rules should be listed only once in the category tree, with an indication that there are multiple rules for this one category (without listing it as two seperate categories, since that's likely to get confusing).

src/util/classes.ts
```diff
@@ -23,26 +24,68 @@ export interface Category {
...
 
-const COLOR_UNCAT = '#CCC';
+const COLOR_GREEN = '#0F0',
+  COLOR_SUPER_GREEN = '#54e346',
+  COLOR_BRIGHT_GREEN = '#A8FC00',
+  COLOR_UNCAT = '#CCC',
+  COLOR_YELLOW = '#FCC400',
+  COLOR_SLIGHTLY_YELLOW = '#f5ff65',
+  COLOR_ORANGE = '#ffba47',
+  COLOR_RED = '#F80';
```
ErikBjare, line 34: Red is orange now? Previous red was `#F33`.

```suggestion
  COLOR_RED = '#F33';
```

src/util/classes.ts
```diff
@@ -23,26 +24,68 @@ export interface Category {
...
 
 // The default categories
 // Should be run through createMissingParents before being used in most cases.
 export const defaultCategories: Category[] = [
   {
-    name: ['Work'],
-    rule: { type: 'regex', regex: 'Google Docs|libreoffice|ReText' },
-    data: { color: '#0F0' },
```
ErikBjare, line 34: Better off keeping this category imo, even if without a rule, such that children inherit color and don't need to be set explicitly for each child.

src/util/classes.ts
ErikBjare:
```suggestion
    data: { color: COLOR_ORANGE },
```

ErikBjare (changes requested):
Was hoping to merge this but I still find the hidden-from-user `select_keys` behavior problematic.

I really want to see a UI for it before I feel okay with adding the selector to the default categories, as I otherwise think users will be very confused.

The "multi-rule" categories also make me uneasy, as I didn't design for it so I would want to review/test this much more thoroughly before I feel confident merging it (see comments for specifics).

**To support multi-rule categories like this without causing trouble, I think it'd be best if we allowed the `rule` property to be a array of rules.** Instead of having several categories with the same name implicitly be the same (troublesome with the `data` property, renaming, parents, etc).

All these color changes are also very hard to review. I generally don't want them changed. I tried to leave comments where I found differences, but I'm sure I didn't spot them all.

If you want, we can strip the `select_keys` and merge this now (so we get the color constants etc.), or we can leave this open if you want to give the UI part a shot.

iloveitaly:
Makes sense! I don't have any additional time to spend on this, so someone else will have to take this up to get it across the finish line.

CVirus:
bump

iloveitaly:
Closing out in favor of the simplier https://github.com/ActivityWatch/aw-webui/pull/495

PR #288, opened 2021-05-07
ci: better and more screenshots (with aw-fakedata)

ErikBjare:


ErikBjare:
Wow, the auto reply with screenshots thing is pretty neat!

@johan-bjareholt Any work done on period_union?

ErikBjare:
Looks like there are some issues with the fakedata generated. 

Needs some more work to look right and work with aw-server-rust, but this is otherwise ready to merge.

ErikBjare:
It's a bit spammy and not perfect (aw-server-rust master still not working 100% right).

But good enough for now, merging.

PR #287, opened 2021-05-05
Style updates

iloveitaly:
Made a bunch of style updates that were bothering me. Added comments in the commit messages.

image

PR #291, opened 2021-06-10
Style updates (continued)

ErikBjare:
Resuming work on #287


ErikBjare:
I think that resolves all the comments I had on #287. Merging.

Thanks again @iloveitaly! :tada:

PR #306, opened 2021-09-29
dev/server-sync-settings

aaayushsingh:
sync settings to server

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "editor.formatOnSave": true
+}
--- a/src/views/settings/Settings.vue
+++ b/src/views/settings/Settings.vue
@@ -6,15 +6,19 @@ div
 
   hr
 
-  DaystartSettings
+  SettingsSyncSettings(v-bind="{ settingsData }")
 
   hr
 
-  TimelineDurationSettings
+  DaystartSettings(v-bind="{ settingsData }")
 
   hr
 
-  LandingPageSettings
+  TimelineDurationSettings(v-bind="{ settingsData }")
+
+  hr
+
+  LandingPageSettings(v-bind="{ settingsData }")
 
   hr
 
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "editor.formatOnSave": true
+}
```
ErikBjare, line 3: You can omit this, and add `.vscode` to `.gitignore`

src/views/settings/Settings.vue
```diff
@@ -47,6 +54,70 @@ export default {
...
+        if (durationDefault) {
+          // existing user, set sync to false by default
+          localStorage.setItem('syncSettingsToServer', 'false');
+        } else {
+          // new user, keep sync on by default
+          localStorage.setItem('syncSettingsToServer', 'true');
+        }
+      }
+      return;
+    },
```
ErikBjare, line 120: Stuff like this should definitely go in a Vuex module.

src/views/settings/SettingsSyncSettings.vue
```diff
@@ -0,0 +1,63 @@
...
+    //     localStorage.removeItem('newReleaseCheckData');
+    //   }
+    // } else {
+    //   console.info('newReleaseCheckData not found in localStorage');
+    // }
+    // },
+    saveData() {
+      // console.log('updating');
+      this.settingsData.syncSettingsToServer = !this.settingsData.syncSettingsToServer;
+      updateSettingOnServer('syncSettingsToServer', this.settingsData.syncSettingsToServer);
```
ErikBjare, line 56: Probably need to set the keys one-by-one (unless the API allows something else).

Perhaps the settings keys should be prefixed with `setting` or `config`, such that `startOfDay` becomes `config.startOfDay` (just in case we want to use the key-value store for something else as well).

ErikBjare (commented):
Good progress so far, looks like next step is putting it in Vuex and start to refactor some uses of `localStorage` in the code.

PR #309, opened 2021-11-14
Set delay to 0 for faster timeline tooltips

GabLeRoux:
Closes following feature request:  
https://forum.activitywatch.net/t/make-tooltip-hover-in-timeline-view-show-instantly/1560

Before:

before

After:

after


ErikBjare:
Nice, thanks!

PR #312, opened 2021-12-09
Started working on Electron-version of ActivityWatch

ErikBjare:
## TODO

 - [x] Get working in dev/serve mode
 - [ ] Proper CLI (using which lib?)
 - [ ] Fix tray icon menu for modules (update the checkboxes' running status)
 - [ ] Bundle servers and watchers (so that they work in the AppImage)

## Before (beta) release

 - [ ] Test on multiple OSs

### DIFF
--- a/.github/workflows/electron.yml
+++ b/.github/workflows/electron.yml
@@ -1,0 +1,81 @@
+name: Build Electron
+
+on:
+  push:
+    branches: [ master ]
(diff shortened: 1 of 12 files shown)


ErikBjare:
This is never happening. But the dream lives on in https://github.com/ActivityWatch/aw-tauri

PR #314, opened 2022-01-05
fix: keep shown currentView when date or periodLength changes

ErikBjare:
Reported by @NicoWeio


ErikBjare:
Hah, the screenshot revealed a bug...

image

Probably the first time the screenshots revealed a bug like this :)

PR #320, opened 2022-03-02
feat: added ability to edit events from timeline view

ErikBjare:
Requires the following to fetch single events by ID (needed for editing events cut off by query):

 - [x] https://github.com/ActivityWatch/aw-core/pull/109
 - [x] https://github.com/ActivityWatch/aw-server/pull/90
 - [x] https://github.com/ActivityWatch/aw-server-rust/pull/268
 - [x] https://github.com/ActivityWatch/aw-client/commit/32c25fd5a7b6cf78cb06256ee9394eac758128a1
 - [x] https://github.com/ActivityWatch/aw-client-js/pull/36

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,9 @@
 module.exports = {
   presets: ['@vue/cli-plugin-babel/preset'],
-  plugins: ['lodash', '@babel/plugin-syntax-dynamic-import'],
+  plugins: [
+    'lodash',
+    '@babel/plugin-syntax-dynamic-import',
+    '@babel/plugin-proposal-nullish-coalescing-operator',
+  ],
   comments: false,
 };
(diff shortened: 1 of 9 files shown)


johan-bjareholt:
Can this be an issue when editing events at the end/start of the current start time/endtime. Not sure if it applies to both aw-server-python and aw-server-rust, but I'm quite certain that it at least applies for aw-server-rust since it cuts the starttime or duration if it is outside starttime or endtime to give more accurate results.

johan-bjareholt:
That's also the reason why we didn't end up adding support for editing in the search view.

ErikBjare:
@johan-bjareholt Ah, yeah that's right. I'll look into a way to get the full event to be edited.

ErikBjare:
@johan-bjareholt I've now added a API to aw-server and aw-server-rust to fetch a single event by its ID, and will use that to fetch the full event to be edited.

PR #322, opened 2022-03-03
VisTimeline: hide buckets without events in specified time range

NicoWeio:
I finally got to implement a little feature I suggested on Discord a month earlier:
> Buckets without events in the specified time range shouldn't appear in the Timeline view. IMO all they do is take up vertical space…

I also followed @ErikBjare's recommendation and added an info box for when this happens:
> X buckets are hidden because they had no events in the specified range

This PR is work-in-progress as I'm still trying to clean up my code.

### DIFF
--- a/src/visualizations/VisTimeline.vue
+++ b/src/visualizations/VisTimeline.vue
@@ -1,6 +1,8 @@
 <template lang="pug">
 div
   div#visualization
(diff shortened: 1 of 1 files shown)


ErikBjare:
Woopsie, I just merged a thing just like this in: https://github.com/ActivityWatch/aw-webui/pull/320/files

PR #323, opened 2022-03-03
fix: check for missing ID in New View modal

prplecake:
Mitigates activitywatch/activitywatch#706

If there's any sort of migration or data integrity checks run when upgrading AW, maybe it should look for views with a null ID and set the ID to a slugified view name?


ErikBjare:
Thanks!

>If there's any sort of migration or data integrity checks run when upgrading AW, maybe it should look for views with a null ID and set the ID to a slugified view name?

Not yet. This will have to do for now :)

PR #327, opened 2022-03-15
refactor: settings store

ErikBjare:
- fix #600 timline cut off
- style: cleanup not needed changes
- add syncsettings, add server api call on settings change
- refactor: refactored settings in preparation for server-side storage of settings


ErikBjare:
Might be a few bugs hiding in here, but merging for now as it's such a big improvement.

PR #328, opened 2022-03-15
Heatmap viz disappears in show last when selecting something bigger than 1h

harigl:
https://trello.com/c/MK3SZH7D/17-heatmap-viz-disappears-in-show-last-when-selecting-something-bigger-than-1h


ErikBjare:
I think you made this PR against the wrong repo... :)

PR #334, opened 2022-04-18
feat: added "Ctrl+Enter" shortcut for querying in the Query Explorer

NicoWeio:
It just feels natural… 😄

I also added a seemingly missing comma. I'm not familiar with _pug_, though, so this may be wrong.


ErikBjare:
Very nice, thanks!

PR #344, opened 2022-06-08
feat: add update button to Timeline view (InputTimeInterval)

### LINKED ISSUE #267: Refresh timeline view button
Feature request - I love the time line view, but wish there was a faster way to see new entries than changing the duration to something else then back to make it refresh, or to change the date, press update, and see new timeline content.

Can there just be one update button that can always be clicked to refresh the timeline?

Furffico:
Closes #267

I am new to ActivityWatch and found that I do need a refresh button on Timeline, so I implemented it. (Bucket view is also affected.)
The update button is visible only when interval mode is *last duration* and prop `showUpdate` is true. A label to show the time of last updating is also added beneath the button.


src/components/InputTimeInterval.vue
```diff
@@ -38,13 +38,24 @@ div
             :disabled="mode == 'range' && (invalidDaterange || emptyDaterange || daterangeTooLong)",
             @click="valueChanged"
           ) Update
+      
+    div(style="float:right;text-align:right" v-if="showUpdate && mode=='last_duration'")
```
ErikBjare, line 42: Avoid float, use flexbox instead. You can use these helper classes: https://getbootstrap.com/docs/4.0/utilities/flex/

src/components/InputTimeInterval.vue
```diff
@@ -38,13 +38,24 @@ div
...
+      b-button.px-2(@click="update()", variant="outline-dark")
```
ErikBjare, line 43: Maybe?

```suggestion
      b-button.px-2(@click="update()", variant="outline-dark" size="sm")
```

src/components/InputTimeInterval.vue
```diff
@@ -38,13 +38,24 @@ div
...
+        icon(name="sync")
+        span.d-none.d-md-inline
+          |  Update
+      div.mt-1(style="color: #999" v-if="lastUpdate")
```
ErikBjare, line 47:
```suggestion
      div.mt-1.small(style="color: #999" v-if="lastUpdate")
```

ErikBjare (commented):
Nice! Just a few things :)

src/components/InputTimeInterval.vue
```diff
@@ -95,9 +111,17 @@ export default {
...
         (!this.emptyDaterange && !this.invalidDaterange && !this.daterangeTooLong)
       ) {
+        this.lastUpdate = moment();
         this.$emit('input', this.value);
       }
     },
+    update() {
+      if (this.mode == 'last_duration') {
+        this.mode = ''; // remove cache on v-model
+        this.mode = 'last_duration';
```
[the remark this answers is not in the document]
Furffico: It's kinda wired that the button won't work without line 120. 😣
I guess the cause is that the computed property `value` (line 85) would be cached by Vue. You may refer to [this example from official documentation](https://vuejs.org/guide/essentials/computed.html#computed-caching-vs-methods). 
The documentation suggests to use methods instead, which is not avaliable for our situation, since it is difficult to call the methods from parent components.
I also tried to use a function (like `()=>[moment().subtract(this.duration, 'seconds'), moment()]`) as the return value of the getter, but this will disable the watchers in Timeline and make things more complex.
So I figured out this simple-yet-strange solution: I can change the value of a reactive dependency (`this.mode`) to something else in order to invoke a update on the cached value. And to remain consistency it is changed it back immediately. After `this.update()`, the cached values are updated so the watchers in the parent components would be invoked, resulting a update in the timeline.

src/components/InputTimeInterval.vue
Furffico: Thank you for the suggestions😊

src/components/InputTimeInterval.vue
```diff
@@ -95,9 +111,17 @@ export default {
         this.mode == 'last_duration' ||
         (!this.emptyDaterange && !this.invalidDaterange && !this.daterangeTooLong)
       ) {
+        this.lastUpdate = moment();
         this.$emit('input', this.value);
       }
     },
+    update() {
+      if (this.mode == 'last_duration') {
+        this.mode = ''; // remove cache on v-model
```
ErikBjare, line 120:
```suggestion
        this.mode = ''; // remove cache on v-model, see explanation: https://github.com/ActivityWatch/aw-webui/pull/344/files#r892982094
```

src/components/InputTimeInterval.vue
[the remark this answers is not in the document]
ErikBjare: Ah, I get it now. Thanks for the explanation!

I could probably figure out another way to get around it (make the `value` computed depend on the `lastUpdated` reactive?), but this seems good enough.

ErikBjare:
Merged!

Thank you for your contribution! :star2:

PR #349, opened 2022-07-05
feat: started working on multi-device queries

ErikBjare:
A few TODOs here and there in the comments.

Goal is to get it to a state where it's at least mergeable, but shoud prob hidden behind a toggle or something until it's more robust.

Also stumbled into a weird import bug when moving around some buckets (bucket was created, but no events were inserted). Buckets were exported from aw-server-rust, and imported into aw-server-python (in case it matters).

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -87,7 +87,7 @@ export function canonicalEvents(params: DesktopQueryParams | AndroidQueryParams)
     params.classes ? `events = categorize(events, ${classes_str});` : '',
     // Filter out selected categories
     params.filter_classes ? `events = filter_keyvals(events, "$category", ${cat_filter_str});` : '',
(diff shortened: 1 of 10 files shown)


src/queries.ts
```diff
@@ -86,6 +87,8 @@ export function canonicalEvents(params: DesktopQueryParams | AndroidQueryParams)
     params.classes ? `events = categorize(events, ${classes_str});` : '',
     // Filter out selected categories
     params.filter_classes ? `events = filter_keyvals(events, "$category", ${cat_filter_str});` : '',
+    // "Return" events by setting variable named with return_suffix if set
```
ErikBjare, line 90:
```suggestion
    // "Return" events by setting variable named with return_variable if set
```

PR #361, opened 2022-08-26
add NeoChat matrix client to Comms::IM

hooger:


iloveitaly:
@ErikBjare seems to make sense as well!

PR #364, opened 2022-09-06
ci: fixed if-statement always evaluating to true in comment.yml

JarLob:
The expression was treated as a string and always evaluated as true.


ErikBjare:
Thanks @JarLob!

PR #365, opened 2022-09-07
feat: started working on force-directed graph for visualizing category-transitions

ErikBjare:
Generates a graph where:

 - node radius is based on the time spent in the category.
 - edge force & thickness is based on the number of transitions (not the best metric)

Uses category colors.

## TODO

 - [ ] What should count as a transition/context-switch?
    - Currently, even the briefest of transitions are counted as just as significant (could reduce this by merging/flooding category-events and removing short category-events)
    - Also shouldn't count transitions wider than x seconds apart (or fill such gaps with events that have an "afk" category, since that might also be interesting to see)
    - @casaout might have some ideas
 - [ ] Is there a way to get directed-graph behavior, with animated edges to see direction things are flowing?
    - Seems super overkill and hard, but maybe awesome?

## Screenshot (WIP)

image

### DIFF
--- a/src/visualizations/ForceGraph.vue
+++ b/src/visualizations/ForceGraph.vue
@@ -2,6 +2,12 @@
 div#forcegraph
 </template>
 
+<style>
+#forcegraph > svg {
+  border: 1px solid #555;
+}
+</style>
+
 <script lang="ts">
 import * as d3 from 'd3';
 
@@ -13,23 +19,27 @@ export default {
       required: true,
     },
   },
-
+  data: () => ({
+    cancelPromise: null,
+  }),
   watch: {
     // Watch for changes in the data and update the graph
-    data: {
-      handler: function () {
-        this.drawGraph(this.data);
-      },
-      deep: true,
+    data() {
+      this.drawGraph(this.data);
     },
   },
-
+  mounted() {
+    this.drawGraph(this.data);
+  },
   methods: {
     drawGraph({ nodes, links }) {
       console.log('rendering...');
-      console.log(nodes, links);
-      const svgEl = ForceGraph({ nodes, links }, { nodeTitle: d => d.id });
-      console.log('appending...');
+      this.cancelPromise && this.cancelPromise();
+      const promise = new Promise(resolve => {
+        this.cancelPromise = resolve;
+      });
+
+      const svgEl = ForceGraph({ nodes, links }, { invalidation: promise });
       const svg = d3.select('#forcegraph');
       //clear
       svg.selectAll('*').remove();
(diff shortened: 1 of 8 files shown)


src/visualizations/ForceGraph.vue
```diff
@@ -0,0 +1,210 @@
...
+      required: true,
+    },
+  },
+  data: () => ({
+    cancelPromise: null,
+  }),
+  watch: {
+    // Watch for changes in the data and update the graph
+    data() {
+      console.log('data changed');
```
ErikBjare, line 28:
```suggestion
```

casaout:
Thanks, Erik, for pinging me!

It might be helpful to get a little bit more context on the feature aims and what you're trying to achieve/visualize, to be able to answer your question better. 

Am I assuming correctly that you create categories based on the app/website usage that ActivityWatch detects?

Generally, I think it might be really challenging to come up with good, universal rules/heuristics that define which category switches (or combinations) account for a context switch. It's also highly dependent on the definition of 'context switch'; I'd probably define it to something that is similar to a task (i.e., task switch)...  In some preliminary work, we've [tried to train models ](https://andre-meyer.ch/DATA/Pubs/2020_TaskDetection_PrePrint.pdf) to identify task switches and types based on users' self-reports; but these are highly individual.

What we're doing in [PersonalAnalytics](https://github.com/haSEL-UZH/PersonalAnalytics) is to treat 'afk' as it's own 'IDLE' category. This is set if there is no user input for 2 (or 3) minutes, I think.   The thing is, however, that the user might be taking some notes (but is still within the same context/task) or doing a biobreak (which would probably be a context/task switch).

ErikBjare:
> It might be helpful to get a little bit more context on the feature aims and what you're trying to achieve/visualize, to be able to answer your question better.

I'm not quite sure yet, this is quite experimental. I wanted to try and visualize how I context switch between categories in some neat way.

Asked you because I figured you had a good view of the prior art and could point me in the right direction (since I noticed I was quite lost). Didn't want to reinvent the wheel.

> Am I assuming correctly that you create categories based on the app/website usage that ActivityWatch detects?

Yes, the categorization happens using regular expressions, and form a hierarchy.

> This is set if there is no user input for 2 (or 3) minutes, I think.

We use 3min :)

> The thing is, however, that the user might be taking some notes (but is still within the same context/task) or doing a biobreak (which would probably be a context/task switch).

Yeah, I've figured that some context switches are larger/harder than others. Perhaps one could let the user set a threshold for how "large" switches have to be to count, somehow (maybe depending on gap size, low-likelihood transition, ...).

Will read that paper, thanks!

casaout:
A few things that came to mind:
- The interesting thing is that I haven't been able to find a proper solution in related work and myself yet. So it's definitely awesome to have another great mind thinking about it ;)
- You could also consider the **time spent in an activity (category) before switching** to the next one. I could imagine that you, e.g., switch less when working in the IDE.
- Maybe the transition itself is the most interesting. Are there any **patterns**? (e.g. after social media, a person always switches to email briefly; or, more switches towards the evening when tired)   Could some patterns (e.g. constant switching to email/IM when coding) indicate that you are not very focused/productive/...? 
- Regarding transitions between activities, one thing, we also looked at briefly was to identify patterns of switches that frequently come together (code -> stackoverflow -> code, email -> IM -> task list). We also did some preliminary animated visualizations, but I cannot find them anymore :( Somewhat related, maybe Chapter 5.2 in [our paper on developers' work lives](https://www.zora.uzh.ch/id/eprint/136503/1/productiveWorkday_TSE17.pdf) could be interesting too.
- Generally, I believe such a visualization will benefit from user input, be it manual setting of thresholds/creating simple queries OR self-reporting (productivity/tiredness/focus/progress).

PR #368, opened 2022-09-08
feat: started working on including stopwatch activity in Activity view

ErikBjare:
## TODO

 - [ ] Currently doesn't show up in view because not sharing common keys (`label` instead of `app` and `title`).
    - Should transform the event data into the form: `{app: 'manual', title: '<label>'}`


ErikBjare:
I've hidden it behind a toggle, which in turn is only shown in devmode. 

So, ready to merge. I don't want to have this as yet another dangling PR I'll forget about.

PR #370, opened 2022-09-10
fix: add browser-queries support for waterfox and orion

alclary:
Small edit to appnames list under queries.ts to address issue/request expressed in ActivityWatch/aw-watcher-web#88. I did not add 'Min' browser to the list, as it does not currently have extension support. 

I tested with waterfox on a windows machine and it worked. I am unable to test Orion, as I do not have an OSX machine readily available.

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -222,18 +222,28 @@ const browser_appnames = {
     'Brave-browser',
   ],
   firefox: [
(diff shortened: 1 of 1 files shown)


ErikBjare:
Will eagerly merge this and hope it fixes it for some folks.

Thanks @alclary!

PR #371, opened 2022-10-04
feat: pull default query from query param

iloveitaly:
- Load default query from `q` param in url. This allows users to have bookmarks with various queries.
- Expand default query input area. For my taste, it was a bit too small.


ErikBjare:
Neat, thanks!

PR #373, opened 2022-10-10
docs: better explain how to use pre-releases for dev/testing

ShootingKing-AM:
There is no generated `static` folder in case of aw-server-rust (unlike aw-server-python), even while packaging rust server, files are directly taken from `aw-server-rust/aw-webui/dist/`.

[aw-server-rust/Makefile](https://github.com/ActivityWatch/aw-server-rust/blob/7c2b31f173194d75634079128a27ed06d83365b1/Makefile#L86)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ You can run a development version of aw-webui with your main version of Activity
 The assets are stored in the following directories (relative to your installation directory), depending on if you use aw-server-python (default) or aw-server-rust:
 
  - aw-server-python: `activitywatch/aw-server/aw_server/static/`
- - aw-server-rust: `activitywatch/aw-server-rust/static/`
+ - aw-server-rust: `activitywatch/aw-server-rust/aw-webui/dist/`
 
 Either copy the assets manually, or run `make build` from the `aw-server` parent directory to rebuild and copy the assets for you.
 
--- a/README.md
+++ b/README.md
@@ -44,32 +44,38 @@ npm ci
 npm run build
 ```
 
-### Using a pre-release with your main install
+## Using a pre-release with your main install
 
 **Note:** Running a development version of aw-webui with an old aw-server can lead to issues due to version incompatibilities.
+
+### By copying the web-assets to your main install
+---
 
 You can run a development version of aw-webui with your main version of ActivityWatch by building it (or fetching the latest build from CI) and replacing placing the contents of the `static` directory of your aw-server (or aw-server-rust) installation. For simplicity, back up the original directory for easier switching back.
 
 The assets are stored in the following directories (relative to your installation directory), depending on if you use aw-server-python (default) or aw-server-rust:
 
  - aw-server-python: `activitywatch/aw-server/aw_server/static/`
- - aw-server-rust: `activitywatch/aw-server-rust/aw-webui/dist/`
+ - aw-server-rust: `activitywatch/aw-server-rust/static/`
 
-Either copy the assets manually, or run `make build` from the `aw-server` parent directory to rebuild and copy the assets for you.
+You can copy the assets manually from your `make build` or `npm run build` output to the above locations.
 
 Once you've put the files in the directories, you may have to do a hard refresh in your browser to invalidate any stale caches.
 
-If you want to actively iterate on aw-webui with your local production data, you'll want to use a development build, automatically update it, and connect a aw-server running against production data. To do this, in one terminal window run:
+### Using your main install's data
+---
+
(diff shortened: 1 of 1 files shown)


ErikBjare:
Files are only taken from `aw-webui/dist/` if run from source. In a normal installation, the files will indeed be under `aw-server-rust/static` in the installation directory (`/opt/activitywatch/aw-server-rust/static` exists for my installation).

If I've misunderstood, do let me know. Closing for now.

ShootingKing-AM:
The documentation is a little confusing, i thought it was taking about the soruceTree's aw-server dir when ```...or run make build from the aw-server parent directory to rebuild and copy the assets for you...``` - thought the aw-server directory is the sourceTree's (instead of intended installation aw-server directory. Since running ```make build``` in installation aw-server will not yield anything.

Here sourceTree means the folder where we git clone'd or the folder containing the source files.

When building from source in sourceTree, [makefile of aw-server-python](https://github.com/ActivityWatch/aw-server/blob/1d6d24edd1cef91fb843edc771ee3b7a5e3fc294/Makefile#L5-L18) will create a `static` directory unlike [makefile of aw-server-rust](https://github.com/ActivityWatch/aw-server-rust/blob/7c2b31f173194d75634079128a27ed06d83365b1/Makefile#L20-L27), so i was searching for sourceTree/aw-server-rust/static directory which doesnt exist (for 2 hrs😢)

I still feel that the doc is still confusing, my recommendation,
- Either sync the two makefiles and create a static dir in sourceTree when building from source
- Or, Change readme like

```
Either copy the assets manually, or run make build from the aw-server parent directory to rebuild and copy the assets for you.
```
to
```
Copy the assets manually from `make build` output.
```

ErikBjare:
Ah, I can certainly see how that is confusing.

I think changing around how the source tree folder structure looks so that it looks the same as for the installation isn't really possible due to constraints of packaging (it might be, but it would be a mess to change). There will always be a path disparity, and the current one is good enough.

But I'm all for improving the language of the README to make it clear. 

> Change to: Copy the assets manually from `make build` output.

I think this suggestion is good. 

Could you update this PR with this and maybe other changes that would clarify/simplify it? I'll take a second pass at it when you're done :)

ShootingKing-AM:
Updated with some restructuring too :)

README.md
```diff
@@ -44,32 +44,38 @@ npm ci
...
  - aw-server-rust: `activitywatch/aw-server-rust/static/`
 
-Either copy the assets manually, or run `make build` from the `aw-server` parent directory to rebuild and copy the assets for you.
+You can copy the assets manually from your `make build` or `npm run build` output to the above locations.
 
 Once you've put the files in the directories, you may have to do a hard refresh in your browser to invalidate any stale caches.
 
-If you want to actively iterate on aw-webui with your local production data, you'll want to use a development build, automatically update it, and connect a aw-server running against production data. To do this, in one terminal window run:
+### Using your main install's data
+---
```
ErikBjare, line 66:
```suggestion
```

README.md
```diff
@@ -44,32 +44,38 @@ npm ci
 npm run build
 ```
 
-### Using a pre-release with your main install
+## Using a pre-release with your main install
 
 **Note:** Running a development version of aw-webui with an old aw-server can lead to issues due to version incompatibilities.
 
+### By copying the web-assets to your main install
+---
```
ErikBjare, line 52:
```suggestion
```

README.md
```diff
@@ -44,32 +44,38 @@ npm ci
...
 ```
 
-If you want to add `debugger` statements in your code and otherwise break linting rules, you'll need to add a `--skip-plugins=no-debugger` to that command. Then,  in another terminal (with your venv activated, assuming you are using python aw-server) run:
+If you want to add `debugger` statements in your code and otherwise break linting rules, you'll need to add a `--skip-plugins=no-debugger` to that command. 
+Then, in another terminal (with your venv activated) run:
 
 ```shell
-poetry install
-aw-server
+serve -l tcp://localhost:27180 -d ../aw-server/static
```
ErikBjare, line 78: What is this command? I don't have it installed on either my macOS or Linux machine.

Instead, I suggest:

```suggestion
python3 -m http.server --bind 127.0.0.1 27180 --directory ../aw-server/static
```

ErikBjare (approved):
Nice, just a few things.

ShootingKing-AM:
Thanks for the quick review and merge! 🎉

`serve` is Vue's own recommended way of http server. But both do the same thing, py httpd should also do the task. Thanks !

ref: https://cli.vuejs.org/guide/deployment.html#previewing-locally

PR #374, opened 2022-10-10
fix: fixed NewReleaseNotification couldn't be hidden/disabled

ShootingKing-AM:
- Typo is settingStore data name
- component NewReleaseNotification cannot be permanently disabled or hidden since state data is not in sync with settingStore's state

https://github.com/ActivityWatch/aw-webui/blob/1980a403dfc2a9f54c0700c202d20e17ccf8bba6/src/stores/settings.ts#L43-L49

https://github.com/ActivityWatch/aw-webui/blob/1980a403dfc2a9f54c0700c202d20e17ccf8bba6/src/components/NewReleaseNotification.vue#L42-L44

### DIFF
--- a/src/components/NewReleaseNotification.vue
+++ b/src/components/NewReleaseNotification.vue
@@ -40,7 +40,7 @@ export default {
     };
   },
   computed: {
-    ...mapWritableState(useSettingsStore, { data: 'newReleaseNotification' }),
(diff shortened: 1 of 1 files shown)


ErikBjare:
Thanks for this @ShootingKing-AM! ❤️

PR #375, opened 2022-10-11
feat: added "count-as-active" setting to always count some apps/titles as active

ShootingKing-AM:
- Never treat some app with process names or titles matching a Regex pattern as AFK
- Added a browser setting for the same
- Future todo can be to implement a server query function of the form filter_keyvalues_regex( <events>, "*", <Regex> ) which will match all event's attrs like categorize() (future url support etc.)

Example data,
x2

Pre-Filter setting,
x

Setting,
y1

Post-Filter setting,
y2

Relevant issues: https://github.com/ActivityWatch/activitywatch/issues/261


src/queries.ts
```diff
@@ -102,6 +104,7 @@ function isMultiParams(object: any): object is MultiQueryParams {
 export function canonicalEvents(params: DesktopQueryParams | AndroidQueryParams): string {
   // Needs escaping for regex patterns like '\w' to work (JSON.stringify adds extra unecessary escaping)
   const categories_str = JSON.stringify(params.categories).replace(/\\\\/g, '\\');
+  const neverTreatAsAfkFilter_str = isDesktopParams(params)? JSON.stringify(params.neverTreatAsAfkFilter).replace(/\\\\/g, '\\'): undefined;
```
ErikBjare, line 107: Why `JSON.stringify` here? Seems unnecessary as it's already a string?

src/queries.ts
```diff
@@ -39,6 +39,7 @@ interface DesktopQueryParams extends BaseQueryParams {
   bid_window: string;
   bid_afk: string;
   filter_afk: boolean;
+  neverTreatAsAfkFilter: string;
```
ErikBjare, line 42: I think `always_active_pattern`, `filter_always_active`, `filter_assume_active` or something similar is a better name (5 word names is a bit much).

src/stores/settings.ts
```diff
@@ -49,6 +50,10 @@ export const useSettingsStore = defineStore('settings', {
     },
     userSatisfactionPollData: {},
 
+    ActivityData: {
+      neverTreatAsAfkFilter: '',
+    },
```
ErikBjare, line 55: I'd rather you put this directly in the top-level of the settings store instead of wrapping in `ActivityData`.

src/views/settings/NeverAfkFilterSettings.vue
```diff
@@ -0,0 +1,35 @@
+<template lang="pug">
+div
+  div.d-sm-flex.justify-content-between
+    div
+      h5.mb-2.mb-sm-0 Never Treat as Afk Filter
+    div
+      b-form-input(size="sm" v-model="NeverAfkFilterSettings")
```
ErikBjare, line 7: Can we add some validation here? Like, don't allow empty or only-whitespace strings (`s.trim().length != 0`) which would otherwise match all/many events accidentally.

ErikBjare (commented):
Very nice! Thank you so much for this!

Just some comments, eager to merge soon :)

src/queries.ts
ShootingKing-AM: my bad, yes its already a string. Was lazy and duplicating the previous line.

src/queries.ts
ShootingKing-AM: modified the 5 word'ed var name. Done.

src/views/settings/NeverAfkFilterSettings.vue
ShootingKing-AM: I didnt want to just modify the setting onChange(), that would require user to click somewhere outside the `input` element to get the setting updated. But yes added some validation.

src/views/settings/NeverAfkFilterSettings.vue
```diff
@@ -0,0 +1,35 @@
...
+          
+  small
+    | Apps or Titles passing this Regex filter will never be counted as afk.
```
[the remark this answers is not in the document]
ShootingKing-AM: I am not that good with consumer-friendly articulation. Did added your suggestion. (Dont know if its enough ?)

ErikBjare:
Added some validation and did some testing, works great now!

Thank you for your contribution @ShootingKing-AM ❤️ 
This feature was a long-time request, happy you stepped up to implement it! ⭐

src/queries.ts
```diff
@@ -102,6 +104,9 @@ function isMultiParams(object: any): object is MultiQueryParams {
 export function canonicalEvents(params: DesktopQueryParams | AndroidQueryParams): string {
   // Needs escaping for regex patterns like '\w' to work (JSON.stringify adds extra unecessary escaping)
   const categories_str = JSON.stringify(params.categories).replace(/\\\\/g, '\\');
+  const always_active_pattern_str = isDesktopParams(params)
+    ? params.always_active_pattern.replace(/\\\\/g, '\\')
```
ErikBjare, line 108: Oops, this replace is probably bad given it's not using `JSON.stringify`. Will fix that in commit directly to master.

ShootingKing-AM: That's embarrassing xD Thanks for the fix :3

dtgay:
Seems like this feature doesn't work on Brave due to https://github.com/ActivityWatch/activitywatch/issues/461. Can anyone else confirm?

Looks like https://github.com/ActivityWatch/aw-webui/pull/375 might be a workaround but doesn't appear available in latest (0.12.1).

PR #380, opened 2022-10-26
feat: Chronological timeline visualization

ShootingKing-AM:
- Shows chronological events on a  timeline for currentDay ± 1day from all (non-empty) buckets.
- For now, does not support greater timeperiod than 'day'
- Automatically updates the data with date-period change
- To add this Visualization, Activity page -> "Edit View" -> "Add Visualization". Click on the Visualization's Settings (⚙) -> "Daily Timeline"
- PR Also fixes large visualizations not being displaced as col-lg-12

image

### DIFF
--- a/src/components/SelectableVisualization.vue
+++ b/src/components/SelectableVisualization.vue
@@ -76,6 +76,8 @@ div
       aw-sunburst-clock(:date="date", :afkBucketId="activityStore.buckets.afk[0]", :windowBucketId="activityStore.buckets.window[0]")
     div(v-if="type == 'custom_vis'")
       aw-custom-vis(:visname="props.visname" :title="props.title")
+    div(v-if="type == 'vis_timeline' && isSingleDay")
+      vis-timeline(:buckets="timeline_buckets", :showRowLabels='true', :queriedInterval="timeline_daterange")
 </template>
 
 <style lang="scss">
@@ -103,6 +105,9 @@ import { build_category_hierarchy } from '~/util/classes';
 
 import { useActivityStore } from '~/stores/activity';
 import { useCategoryStore } from '~/stores/categories';
+import { useBucketsStore } from '~/stores/buckets';
+
+import moment from 'moment';
 
 function pick_subname_as_name(c) {
   c.name = c.subname;
@@ -137,6 +142,7 @@ export default {
         'timeline_barchart',
         'sunburst_clock',
         'custom_vis',
+        'vis_timeline',
       ],
       // TODO: Move this function somewhere else
       top_editor_files_namefunc: e => {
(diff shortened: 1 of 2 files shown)


ShootingKing-AM:
Erik, was not able to get data from fullDesktopQueries itself :( 
- had to get current day +- 1 day events, else it seems abrupt while using day-to-day basis
- had to get all buckets
- getting events, have to package them so that VisTimeline can consume it
- its getting messier

This PR seems a lot more cleaner, maintainer-friendly(D:) implementation

src/components/SelectableVisualization.vue
```diff
@@ -259,9 +270,39 @@ export default {
...
     },
   },
+  watch: {
+    timeline_daterange: async function () {
+      await this.getTimelineBuckets();
+    },
+  },
+  mounted: async function () {
+    await this.getTimelineBuckets();
+  },
```
ErikBjare, line 293: This should only happen if the vis_timeline component is actually used, and then only once for a given set of view parameters and not once-per-`SelectableVisualization` component (currently prob spams requests?).

Might want to write a small wrapper around vis-timeline to do the needed fetching there? Or some other solution, haven't thought about it too deeply.

ErikBjare (commented):
Very neat little PR, love the component reuse! 🌟 

Just have some performance concerns.

src/components/SelectableVisualization.vue
ShootingKing-AM: Nice catch ! I can "type" check the current visualization then conditionally getBuckets ?! Next update occur in computed daterange change (view params changes as expected)

Since vis-timeline is already core component, didn't write a wrapper around it :3

ErikBjare: Ah yes ofc that could work. Still a kind of messy place to put the code in, but I could fix that later.

src/components/SelectableVisualization.vue
```diff
@@ -259,9 +270,44 @@ export default {
...
+
+      // Get events data from CurrentData ± 1 Day so that there is some continuity
+      return [moment(date).subtract(1, 'day'), moment(date).add(1, 'day')];
+    },
     isSingleDay: function () {
       return _.isEqual(this.activityStore.query_options.timeperiod.length, [1, 'day']);
     },
   },
+  watch: {
+    timeline_daterange: async function () {
```
ErikBjare, line 287: This is not in data or a computed? Dead code?

ShootingKing-AM: if you mean timeline_daterange, its there as computed on line 273

ErikBjare: Ahh sorry, my bad, I thought it was a method.

ErikBjare:
Going to ahead and merge this, will test it next week :)

Thanks for the contribution @ShootingKing-AM! ❤️

ShootingKing-AM:
Thanks for the very quick review and merge :3

PR #394, opened 2023-01-17
Fix README and updated gitignore

fbjorn:
I tried to debug Web UI locally, but realised that instructions had a typo.
Also added IDEA to gitignore for anyone using JetBrains IDEs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ test/e2e/reports
 # editor-specific
 *.swp
 .vscode
+.idea
 
 # python
 __pycache__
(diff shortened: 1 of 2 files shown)


fbjorn:
Oh BTW I wasn't able to follow the instructions in the end. Because you need to configure CORS on "main" backend for that, and I didn't understand how should I modify `aw-server.toml` in my installed application  (I'm on macOS)

Instead, I ran another server and it worked:
```bash
AW_SERVER_URL="'http://localhost:5601'" npm run build -- --watch --dest=../aw_server/static
# in another tab:
cd ..
poetry run aw-server --port 5601
```

Not sure this method worth mentioning and how correct it is

PR #490, opened 2023-10-05
fix: support Yandex browser on macOS

750:
This is a popular Chromium-based browser in Russia
`CFBundleDisplayName` is `Yandex` for me on MacOS for app∂ version 23.9.0.2296


ErikBjare:
Thanks!

PR #501, opened 2023-11-15
feat: added support for query caching in aw-client

ErikBjare:
Builds on https://github.com/ActivityWatch/aw-client-js/pull/44

Also fixed so that we don't fetch bucket events if vis_timeline is not in the Activity view. (big perf improvement, major decrease in requests!)

For now there will be some verbose logs to make it easy to debug if something is not working, but it can easily be removed in the future.

For now the cache is cleared on page reload.

### DIFF
--- a/src/components/SelectableVisualization.vue
+++ b/src/components/SelectableVisualization.vue
@@ -306,6 +306,7 @@ export default {
   },
   methods: {
     getTimelineBuckets: async function () {
+      if (this.type != 'vis_timeline') return;
       if (!this.timeline_daterange) return;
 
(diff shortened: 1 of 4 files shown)


src/components/SelectableVisualization.vue
```diff
@@ -306,6 +306,7 @@ export default {
   },
   methods: {
     getTimelineBuckets: async function () {
+      if (this.type != 'vis_timeline') return;
```
ErikBjare, line 309: This got rid of a ton of unnecessary requests!

PR #510, opened 2023-11-21
add field to customize font

Rickard-Martensson:
Currently the only way to modify the default font is to build it yourself. I feel like its a basic enough feature to give it its own field in the settings

### DIFF
--- a/src/style/style.scss
+++ b/src/style/style.scss
@@ -1,17 +1,37 @@
 @import "globals";
 
+
+
+
+
+$default-font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Helvetica Neue", "Ubuntu", sans-serif;
+$header-font-family: 'Varela Round', sans-serif;
+$canvas-font-family: sans-serif;
+
+:root {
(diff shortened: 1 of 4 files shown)


ErikBjare:
Sorry, but I do not think this is important enough to make it a whole section in the settings. Almost no one will change this setting, so I don't want to maintain it.

If you want to change the font, you can use a browser extension/user script to override. You don't need to rebuild the thing to change it that way.

But since I appreciate the effort, I can still merge the part where you refactored the style.css, so you get some contributor cred.

src/style/style.scss
```diff
@@ -1,17 +1,37 @@
 @import "globals";
 
+
+
+
+
+$default-font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Helvetica Neue", "Ubuntu", sans-serif;
+$header-font-family: 'Varela Round', sans-serif;
+$canvas-font-family: sans-serif;
```
ErikBjare, line 9: Not sure what's different for canvas vs default? 
Perhaps they should be combined?

PR #515, opened 2023-11-28
feat: continued work on Calendar visualization

ErikBjare:


TimeToBuildBob:
## 🤖 AI code review

Calendar.vue replaces bucket selection with host selection, fetches events via a canonicalEvents query, and merges adjacent events by category. It adds a mergeAdjacent helper, a loadEventsCanonical method, and a watch on selectedHost. queries.ts exports canonicalEvents in the default export.

### Not safe to merge — 2 P1 open

Confidence 2/5

<details><summary><b>3 findings</b> · ❌ 2 P1 · ⚠️ 1 P2</summary>

**❌ P1 high** — `src/visualizations/Calendar.vue:163`

mergeAdjacent assumes events are in ascending timestamp order (it checks if the next event starts within 10 seconds after the previous event's end), but loadEventsCanonical sorts the query results descending with `_.orderBy(data[0], ['timestamp'], ['desc'])`. With descending order, the compared diff is always negative (because event.timestamp is earlier than lastEvent.timestamp + duration), so the isAdjacent condition is always true for any same-category pair, regardless of the actual time gap. This causes every same-category event in the week to be merged into a single event spanning from the last (earliest) to the first (latest) with a duration that is the sum of all individual durations, missing the gaps between them. The calendar will show one huge event per category instead of discrete periods of activity.

```suggestion
events = _.orderBy(data[0], ['timestamp'], ['asc']);
```

<sub>**How this was verified:** Traced the logic: mergeAdjacent at lines 37-41 computes moment(event.timestamp).diff(lastEnd); with descending order event.timestamp &lt; lastEnd, so the diff is negative, thus &lt; 10 always. Confirmed the sort on line 163.</sub>

**❌ P1 high** — `src/visualizations/Calendar.vue:70`

The initial `selectedHost` is hardcoded to `'erb-m2.localdomain'` rather than derived from the available buckets. The `hostnames` computed property builds the list from `this.buckets`, but the default value is independent of that list. On any machine whose hostname is not `'erb-m2.localdomain'`, the first `loadEventsCanonical` call in `mounted` constructs bucket IDs like `aw-watcher-window_erb-m2.localdomain`, which do not exist, so the que​ry returns no events and the calendar is empty until the user manually selects their real hostname from the dropdown. This is a functional defect for all users other than the developer's machine.

```suggestion
selectedHost: this.hostnames[0] || null,
```

<sub>**How this was verified:** Checked that selectedHost is only set here and not initialized from hostnames in data(); mounted calls loadEventsCanonical which uses this.selectedHost to build bid_window/bid_afk.</sub>

**⚠️ P2 medium** — `src/visualizations/Calendar.vue:155`

The query builder splits the generated query on every semicolon and adds a semicolon back without filtering empty segments. Since the query string ends with a semicolon and then `RETURN = events;` is appended, the final split produces a trailing empty string, which becomes a statement containing just a semicolon. Many query engines (including the aw query parser) reject empty statements or interpret them as a syntax error, which would cause `this.$aw.query` to throw, leaving `this.events` still [] and the calendar blank. The existing `querystr_to_array` in queries.ts filters out empty strings before adding the semicolon, but this code does not reuse it.

```suggestion
query = query.split(';').map(s => s.trim()).filter(s => s).map(s => s + ';');
```

<sub>**How this was verified:** The query string built by canonicalEvents ends with a semicolon (see queries.ts line 161). Appending 'RETURN = events;' adds another semicolon. Splitting on ';' yields a trailing empty string. The code does not filter empty segments, unlike querystr_to_array which filters with .filter(s =&gt; s).</sub>

</details>

<details><summary><b>Files changed (2)</b> — the diff as I read it</summary>

- `src/queries.ts` — Adds canonicalEvents to the default export object.
- `src/visualizations/Calendar.vue` — Rebuilds event loading: host selector, canonical query, mergeAdjacent, and async loading in mounted and selectedHost watcher.

</details>

<sub>Reviewed `530818948632` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 139s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

src/visualizations/Calendar.vue
```diff
@@ -67,37 +108,73 @@ export default {
...
+      console.log(query);
+      query = query.split(';').map(s => s.trim() + ';');
+
+      const timeperiods = [
+        moment(this.queryOptions.start).format() + '/' + moment(this.queryOptions.stop).format(),
+      ];
+      console.log('Querying');
+      const data = await this.$aw.query(timeperiods, query);
+      console.log(data);
+      let events = _.orderBy(data[0], ['timestamp'], ['desc']);
```
TimeToBuildBob, line 163: ❌ **P1** — mergeAdjacent assumes events are in ascending timestamp order (it checks if the next event starts within 10 seconds after the previous event's end), but loadEventsCanonical sorts the query results descending with `_.orderBy(data[0], ['timestamp'], ['desc'])`. With descending order, the compared diff is always negative (because event.timestamp is earlier than lastEvent.timestamp + duration), so the isAdjacent condition is always true for any same-category pair, regardless of the actual time gap. This causes every same-category event in the week to be merged into a single event spanning from the last (earliest) to the first (latest) with a duration that is the sum of all individual durations, missing the gaps between them. The calendar will show one huge event per category instead of discrete periods of activity.

src/visualizations/Calendar.vue
```diff
@@ -15,26 +15,67 @@ div.mx-3
...
   },
   props: {
     buckets: { type: Array },
   },
   data() {
-    return { fitToActive: false, selectedBucket: null, view: 'timeGridDay' };
+    return {
+      events: [],
+      fitToActive: false,
+      selectedHost: 'erb-m2.localdomain',
```
TimeToBuildBob, line 70: ❌ **P1** — The initial `selectedHost` is hardcoded to `'erb-m2.localdomain'` rather than derived from the available buckets. The `hostnames` computed property builds the list from `this.buckets`, but the default value is independent of that list. On any machine whose hostname is not `'erb-m2.localdomain'`, the first `loadEventsCanonical` call in `mounted` constructs bucket IDs like `aw-watcher-window_erb-m2.localdomain`, which do not exist, so the que​ry returns no events and the calendar is empty until the user manually selects their real hostname from the dropdown. This is a functional defect for all users other than the developer's machine.

src/visualizations/Calendar.vue
```diff
@@ -67,37 +108,73 @@ export default {
...
+
+      let query = queries.canonicalEvents({
+        bid_window: 'aw-watcher-window_' + this.queryOptions.hostname,
+        bid_afk: 'aw-watcher-afk_' + this.queryOptions.hostname,
+        filter_afk: this.queryOptions.filter_afk,
+        categories,
+      });
+      query += 'RETURN = events;';
+      console.log(query);
+      query = query.split(';').map(s => s.trim() + ';');
```
TimeToBuildBob, line 155: ⚠️ **P2** — The query builder splits the generated query on every semicolon and adds a semicolon back without filtering empty segments. Since the query string ends with a semicolon and then `RETURN = events;` is appended, the final split produces a trailing empty string, which becomes a statement containing just a semicolon. Many query engines (including the aw query parser) reject empty statements or interpret them as a syntax error, which would cause `this.$aw.query` to throw, leaving `this.events` still [] and the calendar blank. The existing `querystr_to_array` in queries.ts filters out empty strings before adding the semicolon, but this code does not reuse it.

PR #520, opened 2023-12-11
Update app name for Vivaldi browser

vieteh:
On MacOS, the Vivaldi browser app name is only "Vivaldi". Therefore, the UI cannot filter the browser event for Vivaldi correctly.


ErikBjare:
Thanks!

PR #532, opened 2024-01-05
fix: attempt at fixing CSV exports crashing aw-android

ErikBjare:
Untested.

Depends on: https://github.com/ActivityWatch/aw-android/pull/109

### DIFF
--- a/src/views/Buckets.vue
+++ b/src/views/Buckets.vue
@@ -267,16 +267,25 @@ export default {
         );
       });
       const csv = Papa.unparse(data, { columns, header: true });
-      const blob = new Blob([csv], { type: 'text/csv' });
-      const url = URL.createObjectURL(blob);
-      const link = document.createElement('a');
-      link.href = url;
-      link.download = `aw-events-export-${bucketId}-${new Date()
+      const filename = `aw-events-export-${bucketId}-${new Date()
         .toISOString()
         .substring(0, 10)}.csv`;
-      document.body.appendChild(link);
-      link.click();
-      document.body.removeChild(link);
+
+      // Check if the Android interface is available
+      if (window.Android) {
+        // Send the CSV data to the Android code
+        window.Android.downloadCSV(csv, filename);
+      } else {
+        // Fall back to the original method
+        const blob = new Blob([csv], { type: 'text/csv' });
+        const url = URL.createObjectURL(blob);
+        const link = document.createElement('a');
+        link.href = url;
+        link.download = filename;
+        document.body.appendChild(link);
+        link.click();
+        document.body.removeChild(link);
+      }
     },
   },
 };


TimeToBuildBob:
Cannot rebase: TimeToBuildBob has pull-only access on ActivityWatch/aw-webui (`update-branch` 404), and `dev/fix-csv-exports-crashing-app` is Erik's 2024 branch, 245 commits behind. The export path has since moved to `src/util/export.ts`.

The Android crash this targeted is already handled in-app: aw-android injects `ANDROID_EXPORT_HOOK_JS` to intercept blob+download clicks (the current `downloadFileBrowser` path). Leaving the conflict for the author.

src/views/Buckets.vue
```diff
@@ -267,16 +267,25 @@ export default {
...
         .toISOString()
         .substring(0, 10)}.csv`;
-      document.body.appendChild(link);
-      link.click();
-      document.body.removeChild(link);
+
+      // Check if the Android interface is available
+      if (window.Android) {
+        // Send the CSV data to the Android code
+        window.Android.downloadCSV(csv, filename);
```
TimeToBuildBob, line 277: ❌ **P1** — The new Android branch is selected by a truthiness check on `window.Android`, but the code does not verify that `window.Android.downloadCSV` is actually a function. If the Android WebView exposes a `window.Android` object (e.g., for other bridge methods) but does not yet include `downloadCSV` — which is plausible because the bridge is added by a separate aw-android PR that this PR depends on — then `window.Android.downloadCSV(csv, filename)` throws `TypeError: window.Android.downloadCSV is not a function`. This exception is uncaught inside `export_csv`, so the CSV export silently fails (no download, no user-visible error) on any Android WebView that has the object but not the method. The fallback browser path is never reached because the throw happens before the else branch. The fix is to check `typeof window.Android.downloadCSV === 'function'` before calling it, and fall back to the blob download otherwise.

TimeToBuildBob:
## 🤖 AI code review

The PR modifies the CSV export method in Buckets.vue to detect a window.Android interface and, when present, call window.Android.downloadCSV(csv, filename) instead of the browser download path. The filename is now computed once and reused in both branches. The change is untested and depends on an aw-android PR.

### Not safe to merge — 1 P1 open

Confidence 3/5

<details><summary><b>1 finding</b> · ❌ 1 P1</summary>

**❌ P1 high** — `src/views/Buckets.vue:277`

The new Android branch is selected by a truthiness check on `window.Android`, but the code does not verify that `window.Android.downloadCSV` is actually a function. If the Android WebView exposes a `window.Android` object (e.g., for other bridge methods) but does not yet include `downloadCSV` — which is plausible because the bridge is added by a separate aw-android PR that this PR depends on — then `window.Android.downloadCSV(csv, filename)` throws `TypeError: window.Android.downloadCSV is not a function`. This exception is uncaught inside `export_csv`, so the CSV export silently fails (no download, no user-visible error) on any Android WebView that has the object but not the method. The fallback browser path is never reached because the throw happens before the else branch. The fix is to check `typeof window.Android.downloadCSV === 'function'` before calling it, and fall back to the blob download otherwise.

```suggestion
if (window.Android && typeof window.Android.downloadCSV === 'function') {
```

<sub>**How this was verified:** Checked the surrounding code: the condition at line 275 is `if (window.Android)`, and there is no further check on the method. The else branch at line 278 is only reached when window.Android is falsy. The PR description states the bridge depends on a separate aw-android PR, so the method may be absent in some deployments.</sub>

</details>

<details><summary><b>1 advisory finding (summary-only, not scored)</b></summary>

_These P2 guard, heuristic, trade-off, or documentation claims are retained for judgment without opening review threads._

**⚠️ P2 medium** — `src/views/Buckets.vue`

This is a `fix(...)` PR but no test files are included in the diff. Erik's feedback: 'where is the repro & fixes they are supposed to catch' (gptme#3441), 'that measurement should come with a regression test' (gptme#3446). Add a test that would have caught this bug. (Advisory: Erik merged all such PRs but consistently requested tests.)

```suggestion
Add a test file that reproduces the bug before the fix and passes after it.
```

<sub>**How this was verified:** static preflight: fix-commit + touched-files scan (rule 7)</sub>

</details>

<details><summary><b>Files changed (1)</b> — the diff as I read it</summary>

- `src/views/Buckets.vue` — Adds a window.Android branch to export_csv that delegates CSV download to the Android bridge, with the browser blob-download path as fallback.

</details>

<sub>Reviewed `57676d612b22` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 42s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

PR #559, opened 2024-05-01
build: add vite support

ErikBjare:
Trying to keep the webpack build intact for now, while trying to get a working build with vite.

## TODO

 - [x] Fix broken icons in webpack build
 - [x] Fix tests
    - Still broken locally, but passes in CI
 - [x] Ensure CSP is correct

If we are done with the above, we can merge and continue the following in a new PR:
 - [x] Get vite build working
    - [x] Fix `Failed to resolve entry for package "vue-d3-sunburst". The package may have incorrect main/module/exports specified in its package.json`
       - Fixed in https://github.com/David-Desmaisons/Vue.D3.sunburst/pull/54, using fork for now.
     - [x] Fix styles


ErikBjare (commented):
LGTM!

PR #561, opened 2024-05-06
fix: changed many rules in browser-to-appname map to use case-insensitive regexes

ErikBjare:
Replaces https://github.com/ActivityWatch/aw-webui/pull/283

| :rocket: This description was created by [Ellipsis](https://www.ellipsis.dev) for commit f53f8b8b2afdbf3447b4156bb1d25686ddf0f77a  | 
|--------|

### Summary:
Enhanced the `browser_appnames` mapping in `/src/queries.ts` with improved regex patterns for flexible, case-insensitive browser identification, including updates for `LibreWolf`, `Brave`, and other browsers.

**Key points**:
- Updated `browser_appnames` in `/src/queries.ts` to use regex for flexible, case-insensitive matching.
- Regex patterns handle variations in browser names across platforms and versions.
- Enhanced matching for `LibreWolf` with new variant `librewolf-default`.
- Improved regex patterns for `Brave` and other browsers.

----

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -222,9 +222,6 @@ const browser_appnames = {
 
     // Chromium
     /chromium\w?(browser)?(-chromium)?(.exe)?/i,
-
-    // Brave (should this be merged with the brave entry?)
-    'Brave-browser',
   ],
   firefox: [
     // Firefox
@@ -245,7 +242,7 @@ const browser_appnames = {
     /Waterfox(.exe)?/i,
   ],
   opera: [/opera(.exe)?/i],
-  brave: [/brave(.exe)?/i],
+  brave: [/brave\w?(browser)?(.exe)?/i],
   edge: [
     'msedge.exe', // Windows
     /microsoft edge/i, // macOS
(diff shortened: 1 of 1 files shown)


src/queries.ts
```diff
@@ -213,75 +213,48 @@ export function appQuery(
...
-    'Chromium-browser-chromium',
-    'chromium.exe',
-
-    // Pre-releases
-    'Google-chrome-beta',
-    'Google-chrome-unstable',
+    /chromium\w?(browser)?(-chromium)?(.exe)?/i,
 
     // Brave (should this be merged with the brave entry?)
     'Brave-browser',
```
BelKed, line 227: Just curious — Is there some reason for not moving these lines to the Brave Browser category?

ErikBjare: I don't remember...

Found https://discord.com/channels/755040852727955476/755334543891759194/1237102962241704038

Looks like we should merge: https://github.com/ActivityWatch/aw-webui/pull/283

But something stalled it.

ErikBjare:
Looks like I can't pass regexes? Makes sense if the mapping goes into the query...

ErikBjare:
This will never work.

PR #565, opened 2024-05-07
fix: hopefully actually fixes bug in category builder for 'constructor' word (and other prototype stuff)

ErikBjare:
Follow up on https://github.com/ActivityWatch/aw-webui/pull/564

----

| :rocket: This description was created by [Ellipsis](https://www.ellipsis.dev) for commit b5e9eb076de836d10421de37531a98540c025de5  | 
|--------|

### Summary:
Updated `CategoryBuilder.vue` to use `Map` for storing words, replacing object structure to handle prototype-related issues correctly.

**Key points**:
- Changed data structure for storing words from object to `Map` in `CategoryBuilder.vue`.
- Replaced `in` keyword with `Map` methods (`has`, `get`, `set`) to avoid prototype pollution.
- Ensures words like 'constructor' are handled correctly.

----

### DIFF
--- a/src/views/settings/CategoryBuilder.vue
+++ b/src/views/settings/CategoryBuilder.vue
@@ -227,15 +227,15 @@ export default {
           if (word.length <= 2 || this.ignored_words.includes(word)) {
             continue;
           }
-          if (word in words) {
-            words[word].duration += event.duration;
-            words[word].events.push(event);
+          if (words.has(word)) {
+            words.get(word).duration += event.duration;
+            words.get(word).events.push(event);
           } else {
-            words[word] = {
+            words.set(word, {
               word: word,
               duration: event.duration,
               events: [event],
-            };
+            });
           }
         }
       }


ErikBjare:
@john30 Haven't tested (again), but I think this should be correct now.

john30 (commented):
sorry but still not correct. as commented before, all occasions of this.words need to be checked and most probably adjusted as well.
just try the following in the browser console:
your patch would yield in behaviour such as `m=new Map();m.set('example',1);console.log(Object.keys(m))` resulting in an empty array instead of an array with 'example' as the single element, because the Map can not be used like this.
instead, `new Map().keys()` would have to be used but it returns an iterator (and not an array as expected somewhere in the code) so that needs to be adjusted accordingly

just out of curiosity: why do you push and merge immediately before anyone had the chance to review it?

ErikBjare:
@BelKed fixed in https://github.com/ActivityWatch/aw-webui/pull/569

> just out of curiosity: why do you push and merge immediately before anyone had the chance to review it?

Eagerness, clearly to a fault in this case. Sorry about that.

Issue #395, opened 2023-01-22
Timeline - Date range - "Update" button issues


AlexanderMakarov:
In 0.12.1 version there are multiple issues with "Update" button on "Timeline" tab if use "Interval Mode = Date range". All issues are presented on video https://user-images.githubusercontent.com/2630978/213903761-d295b2f6-b95b-4896-aa8c-5fe36cefe94c.mp4

1. UI sends a bunch of requests with "start = start_date" and "end=end_date" on one click.
2. UI doesn't update interface if received 0 events.
3. If don't change start_date/end_date and press "Update" button then nothing happens in sense of back-end requests (imagine situation when user wants to update events on the timeline for "today" or after "Export Raw Events" operation).
4. (optional and more like an improvement) UI doesn't react on "Update" button presses at all (imagine slow back-end when new events coming with decent delay or case when events on one day totally matches events for another day).

Due to I haven't recorded events on this machine for almost the whole month then it gets nothing for chosen (on video) dates. Actual state is below:
![image](https://user-images.githubusercontent.com/2630978/213904050-d6bf7680-7ef5-44de-8802-a5dabe271173.png)

Machine is Linux Mint on 5.4.0-137-generic core, Firefox 109.0 (64-bit).

PR #610, opened 2024-09-08
Improve timeline usability (less clicks to apply time filters)

ideadapt:
I use the Timeline view at least once a day. 
The number of clicks required to filter the time range started to bug me, so I tried to improve it:

- Remove "mode" select. Display both filter directly.
- Use a list of buttons instead of the dropdown to select "last n seconds" (last_duration).
- Apply selected last_duration on button click. Still require explicit button click for time range filter to apply.
- Notify user if no events match (so the user is not irritated, why the timeline does not re-render, addresses an issue mentioned in #395 )
- Apply bootstrap styles. Small layout / design adjustments to align stuff a little.

Fixes https://github.com/ActivityWatch/aw-webui/issues/395

Before:
before

After:
after

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit d15b9efaf6491eb6f9085c007499ce901f8832cc  | 
|--------|--------|

### Summary:
Improves Timeline view usability by reducing clicks for time filters, replacing dropdowns with buttons, and adding user notifications for no events.

**Key points**:
- `src/components/InputTimeInterval.vue`: Removed "mode" select, displaying both filters directly.
- Replaced dropdown with button list for "last n seconds" selection.
- Immediate application of `last_duration` on button click.
- Explicit button click still required for date range filter.
- Added user notification for no matching events.
- Applied Bootstrap styles for layout adjustments.
- `src/views/Timeline.vue`: Adjusted event count display and added alert for no events.

----

### DIFF
--- a/src/components/InputTimeInterval.vue
+++ b/src/components/InputTimeInterval.vue
@@ -112,13 +112,13 @@ export default {
     this.duration = this.defaultDuration;
     this.valueChanged();
 
-    // We want our lastUpdated text to update every ~3s
+    // We want our lastUpdated text to update every ~500ms
     // We can do this by setting it to null and then the previous value.
     this.lastUpdateTimer = setInterval(() => {
       const _lastUpdate = this.lastUpdate;
       this.lastUpdate = null;
       this.lastUpdate = _lastUpdate;
-    }, 3000);
+    }, 500);
   },
   beforeDestroy() {
     clearInterval(this.lastUpdateTimer);
--- a/src/components/InputTimeInterval.vue
+++ b/src/components/InputTimeInterval.vue
@@ -1,55 +1,64 @@
 <template lang="pug">
 div
   div
-    b-alert(v-if="mode == 'range' && invalidDaterange", variant="warning", show)
+    b-alert(v-if="invalidDaterange", variant="warning", show)
       | The selected date range is invalid. The second date must be greater or equal to the first date.
-    b-alert(v-if="mode == 'range' && daterangeTooLong", variant="warning", show)
+    b-alert(v-if="daterangeTooLong", variant="warning", show)
       | The selected date range is too long. The maximum is {{ maxDuration/(24*60*60) }} days.
 
-  div.d-flex.justify-content-between.align-items-end
-    table
-      tr
-        th.pr-2
-          label(for="mode") Interval mode:
-        td
-          select(id="mode", v-model="mode")
-            option(value='last_duration') Last duration
-            option(value='range') Date range
-      tr(v-if="mode == 'last_duration'")
-        th.pr-2
-          label(for="duration") Show last:
-        td
-          select(id="duration", v-model="duration", @change="valueChanged")
-            option(:value="15*60") 15min
-            option(:value="30*60") 30min
-            option(:value="60*60") 1h
-            option(:value="2*60*60") 2h
-            option(:value="4*60*60") 4h
-            option(:value="6*60*60") 6h
-            option(:value="12*60*60") 12h
-            option(:value="24*60*60") 24h
-      tr(v-if="mode == 'range'")
-        th.pr-2 Range:
-        td
-          input(type="date", v-model="start")
-          input(type="date", v-model="end")
-          button(
-            class="btn btn-outline-dark btn-sm",
-            type="button",
-            :disabled="mode == 'range' && (invalidDaterange || emptyDaterange || daterangeTooLong)",
-            @click="valueChanged"
-          ) Update
+  table
+    tr
+      td.pr-2
+        label.col-form-label Show last:
+      td(colspan=2)
+        .btn-group(role="group")
+            input(type="radio", id="dur1", :value="15*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur1").btn.btn-light.rounded-left &frac14;h
+            input(type="radio", id="dur2", :value="30*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur2").btn.btn-light &frac12;h
+            input(type="radio", id="dur3", :value="1*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur3").btn.btn-light 1h
+            input(type="radio", id="dur4", :value="2*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur4").btn.btn-light 2h
+            input(type="radio", id="dur5", :value="3*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur5").btn.btn-light 3h
+            input(type="radio", id="dur6", :value="4*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur6").btn.btn-light 4h
+            input(type="radio", id="dur7", :value="6*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur7").btn.btn-light 6h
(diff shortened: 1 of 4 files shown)


src/components/InputTimeInterval.vue
```diff
@@ -109,7 +118,7 @@ export default {
       const _lastUpdate = this.lastUpdate;
       this.lastUpdate = null;
       this.lastUpdate = _lastUpdate;
-    }, 1000);
+    }, 3000);
```
ErikBjare, line 121: Idk why the comment says 3s, imo it should happen every ~500ms.

src/components/InputTimeInterval.vue
```diff
@@ -1,55 +1,64 @@
...
-            :disabled="mode == 'range' && (invalidDaterange || emptyDaterange || daterangeTooLong)",
-            @click="valueChanged"
-          ) Update
+  table
+    tr
+      td.pr-2
+        label.col-form-label Show last:
+      td(colspan=2)
+        .btn-group(role="group")
+            input(type="radio", id="dur1", :value="15*60", v-model="duration", @change="applyLastDuration").sr-only
```
ErikBjare, line 15: Why `.sr-only`?

src/components/InputTimeInterval.vue
```diff
@@ -1,55 +1,64 @@
...
+            input(type="radio", id="dur6", :value="4*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur6").btn.btn-light 4h
+            input(type="radio", id="dur7", :value="6*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur7").btn.btn-light 6h
+            input(type="radio", id="dur8", :value="12*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur8").btn.btn-light 12h
+            input(type="radio", id="dur9", :value="24*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur9").btn.btn-light 24h
+            input(type="radio", id="dur10", :value="48*60*60", v-model="duration", @change="applyLastDuration").sr-only
+            label(for="dur10").btn.btn-light 48h
```
ErikBjare, line 34: This looks a bit messy, Claude suggested:

```suggestion
        .btn-group(role="group")
          template(v-for="(duration, index) in [
            { value: 0.25, label: '&frac14;h' },
            { value: 0.5, label: '&frac12;h' },
            { value: 1, label: '1h' },
            { value: 2, label: '2h' },
            { value: 3, label: '3h' },
            { value: 4, label: '4h' },
            { value: 6, label: '6h' },
            { value: 12, label: '12h' },
            { value: 24, label: '24h' },
            { value: 48, label: '48h' }
          ]")
            input(
              type="radio"
              :id="'dur' + (index + 1)"
              :value="duration.value * 60 * 60"
              v-model="duration"
              @change="applyLastDuration"
            ).sr-only
            label(:for="'dur' + (index + 1)" v-html="duration.label").btn.btn-light
```

src/components/InputTimeInterval.vue
ErikBjare, line 121:
```suggestion
    }, 500);
```

src/components/InputTimeInterval.vue
```diff
@@ -1,55 +1,64 @@
...
 div
   div
-    b-alert(v-if="mode == 'range' && invalidDaterange", variant="warning", show)
+    b-alert(v-if="invalidDaterange", variant="warning", show)
       | The selected date range is invalid. The second date must be greater or equal to the first date.
-    b-alert(v-if="mode == 'range' && daterangeTooLong", variant="warning", show)
+    b-alert(v-if="daterangeTooLong", variant="warning", show)
       | The selected date range is too long. The maximum is {{ maxDuration/(24*60*60) }} days.
 
-  div.d-flex.justify-content-between.align-items-end
```
ErikBjare, line 9: Any particular thought behind removing the flex and right-side content?

ErikBjare (commented):
Great changes!

My main concern is how does this look on small screens now? Some of the things in the design that you rightfully questioned (the flex and right-float) were actually done they were for small screens, so they look good in the aw-android app.

src/components/InputTimeInterval.vue
ErikBjare: Nvm, Claude explained.

src/components/InputTimeInterval.vue
ideadapt: By right side content you mean the update button with the sync icon?
I think this button is not required anymore. Since the timeline is updated on duration selection automatically, or by clicking the new "apply" button, which is no right next to the date range inputs.

ideadapt:
> Great changes!
> 
> My main concern i s how does this look on small screens now? Some of the things in the design that you rightfully questioned (the flex and right-float) were actually done they were for small screens, so they look good in the aw-android app.

I wasn't aware of the mobile app requirement. I optimized it for screens >=640px. My tests today showed that it would work down to 540px:

new-ui-mobile

Testing the old UI on mobile showed that, at least for my computer host names, the timeline is not useful on mobile screens. The bucket events start to get visible with >450px:

old-ui-mobile

Are there requirements to support screens smaller than 540px? If so, I could try to shrink the duration selection on small screens.

src/components/InputTimeInterval.vue
ErikBjare: The "apply" button only works on the range, not the time interval, so it's not really functionally equivalent. It was meant to be a way to refresh without reloading the entire page.

ErikBjare:
@ideadapt Yes, the same UI is used for aw-android which runs on phones. The bucket id is a lot shorter on android, mine is `aw-android-test` and it takes up ~30% of the width. Not great, but useable.

We used to support landscape mode, so one could make it work pretty well if one wanted, but there was an issue with full reloads on orientation changes, so we disabled it "temporarily" and thats where its still at: https://github.com/ActivityWatch/aw-android/pull/108

So yeah, still need to support/think about small screens, but I'm okay with things not being perfect (since they're not anyway with the tight timeline).

src/components/InputTimeInterval.vue
ideadapt: Ah I see. I never understood the difference between the two update buttons :) thought they solve the same problem ... But now I understand. I have re-added it again. What do you think?

ideadapt: Ping :)

ideadapt: Hi @ErikBjare did you have time to look at it? Did I miss something? For me, the proposed changes in this PR would help me in daily business, and I hope/think others as well.

src/views/Timeline.vue
```diff
@@ -63,6 +65,8 @@ export default {
       const settingsStore = useSettingsStore();
       return Number(settingsStore.durationDefault);
     },
+    // TODO this does not match the actual vis-timeline.chartData which is rendered in the timeline.
+    //  chartData excludes short events.
```
ErikBjare, line 69:
```suggestion
    // This does not match the chartData which is rendered in the timeline, as chartData excludes short events.
```

ErikBjare:
Again, very nice work! I think I'll take it from here :)

Will be really great to have both this and https://github.com/ActivityWatch/aw-webui/pull/609 in the next release!

src/components/InputTimeInterval.vue
ErikBjare: Getting to it now! Will merge today :)

ErikBjare:
Merged! Thank you again for this great work!

ideadapt:
> Merged! Thank you again for this great work!

Cool! Thank you :)

PR #649, opened 2025-03-01
feat: add quick shift buttons to timeline and autocomplete end date from start

### LINKED ISSUE #596: Timeline: select only one date selects this date
If I want to select a specific day in the time line, I have to select the start and end time. Both dates are the same. 

Create the possibility to set only one of the dates and enable the update button. Filter the data for the selected day

RTnhN:
This change has two different changes. 

1. [On the forum](https://forum.activitywatch.net/t/improving-date-selection-in-timeline/3443/3), someone had a suggestion to be able to shift over the date in the timeline view. I also had this problem. It is also nice to have a today button even though you can just click the 24 hour option above the date pickers. The date shift actually works on the start and end dates independently, so you are able to select a two day range in the date pickers and shift that two day range over by a day or a week.  

~~2. While I was working on this change, I also added a change to resolve. This will fill the end date with the start date if the end date is blank. Worst case, the person will just modify the end date to be the date that they really wanted to select. Best case, they wanted a one day range anyways.~~ 

https://github.com/user-attachments/assets/aa32bdf9-9025-493f-9581-537a284b5b85

----

> [!IMPORTANT]
> Adds quick shift buttons to `InputTimeInterval.vue` for date navigation and preserves query parameters in `Activity.vue`.
> 
>   - **Behavior**:
>     - Adds quick shift buttons in `InputTimeInterval.vue` to move date ranges by -1w, -1d, Today, +1d, +1w.
>     - `setToday()` sets both start and end dates to today.
>     - `shiftDay(days)` shifts start and end dates by specified days.
>   - **Navigation**:
>     - Updates `Activity.vue` to preserve query parameters when navigating between periods using `previousPeriod()` and `nextPeriod()`.
>   - **Misc**:
>     - Removes the feature to autocomplete end date from start date if end date is blank.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ActivityWatch%2Faw-webui&utm_source=github&utm_medium=referral)<sup> for 6295ae703563703a0ff23eaebfbfc1b2866b6b62. You can [customize](https://app.ellipsis.dev/ActivityWatch/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/components/InputTimeInterval.vue
+++ b/src/components/InputTimeInterval.vue
@@ -208,9 +208,9 @@ export default {
       this.duration = 0;
       this.valueChanged();
     },
-    shiftDay(days:number) {
-      let currentStart = this.start ? moment(this.start, 'YYYY-MM-DD') : moment();
-      let currentEnd = this.end ? moment(this.end, 'YYYY-MM-DD') : moment();
+    shiftDay(days: number) {
+      const currentStart = this.start ? moment(this.start, 'YYYY-MM-DD') : moment();
+      const currentEnd = this.end ? moment(this.end, 'YYYY-MM-DD') : moment();
       const newStart = currentStart.add(days, 'days').format('YYYY-MM-DD');
       const newEnd = currentEnd.add(days, 'days').format('YYYY-MM-DD');
       this.start = newStart;
--- a/src/components/InputTimeInterval.vue
+++ b/src/components/InputTimeInterval.vue
@@ -27,17 +27,9 @@ div
         td.pr-2
           label.col-form-label.col-form-label-sm Show from
         td
-          input.form-control.form-control-sm.d-inline-block.p-1(
-            type="date", 
-            v-model="start", 
-            style="height: auto; width: auto;"
-          )
+          input.form-control.form-control-sm.d-inline-block.p-1(type="date", v-model="start", style="height: auto; width: auto;")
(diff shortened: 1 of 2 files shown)


RTnhN:
Note: I need to rethink the automatic end date thing. I realized when you change months in the date picker in chrome, it selects the date which selects the end date. I was hoping that it would only select the date whenever you actually click on a calendar day, but that is not the case for the chrome date picker.  Having the date selected on month change just makes the UX not great.

RTnhN:
This also fixes [this problem](https://forum.activitywatch.net/t/using-date-navigation-arrows-clears-filters/4026) from the forum by preserving the query (filter) part of the url when going forwards or backwards. Selecting the date manually preserves it, so I think it should be okay for the buttons to preserve it too.

RTnhN:
@ErikBjare What do you think of this PR? It solves two issues ([this one](https://forum.activitywatch.net/t/improving-date-selection-in-timeline/3443) and [this one](https://forum.activitywatch.net/t/using-date-navigation-arrows-clears-filters/4026/2)) that people were talking about on the forum. I have been personally using the first part of it (the timeline quick shift buttons) for about 3 months, and I find that it makes the timeline view so much better. 

If you want me to split this into two PRs and/or only fix one of the two problems, that is okay with me. Or if you think that this is not worth the change, that is fine.

Issue #611, opened 2024-09-12
Dark mode overrides border color in timeline 


migueldeoleiros:
When using light mode the border of timeline items has the same color as the item itself, but dark mode has a rule that overrides the color for all borders, making it look dark gray for all elements.

This is specially annoying when you have a lot of small items since you can't see the color and the whole item becomes dark gray.

Light mode:
![2024-09-12_10-49-41](https://github.com/user-attachments/assets/d23f090f-7de0-4f65-b2a0-1a08250cb461)
Current dark mode:
![2024-09-12_10-48-26](https://github.com/user-attachments/assets/174c70da-1e08-4e53-8460-e6802c4784d1)
Intended result (although in this case I disabled all border color changes, ideally the lines in the back would stay with the original dark mode color):
![2024-09-12_10-49-16](https://github.com/user-attachments/assets/2d55ad05-c106-4842-938b-bdc6cb5617d9)

migueldeoleiros:
The fix should be pretty easy. I don't know much about CSS, so if someone else can work on it I would appreciate it :) 
Otherwise I may be able to give it a go

PR #705, opened 2025-09-03
Fix dark mode timeline borders

t-fritsch:
This PR fixes the dark mode's timeline colors as mentioned in #611 

before : 
image

after :
image

The fix itself is not the most beautiful css code, but the source of the issue is a quite old css rule that is too generic (`#visualization *` targets almost everything, this is bad) and that uses the `!important` keyword (this is bad too ! 😄😅) :

https://github.com/ActivityWatch/aw-webui/blame/f38b119f0ad628dd8af6ddfaaaf89700b9290c06/static/dark.css#L173-L177

A nicer solution would be to remove this rule and rework dark mode styling from the ground up but this quick fix should be sufficient for now I hope 🙏

### DIFF
--- a/static/dark.css
+++ b/static/dark.css
@@ -170,7 +170,7 @@ hr {
   background-color: #282c32 !important;
 }
 
-#visualization *, svg,
+#visualization *:not([style*=border]), svg,
 .nav.nav-tabs,
 .btn-outline-secondary {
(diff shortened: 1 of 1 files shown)


BelKed:
We don't have merge permissions, only @ErikBjare does

PR #724, opened 2025-11-11
Warn against semicolons in categorisation rules editor

AbdulGoodLooks:
Workaround to attempt to prevent creating rules that break aw-server like this https://forum.activitywatch.net/t/invalid-syntax-for-value-to-assign-out-of-nowhere/4627/3

----

> [!IMPORTANT]
> Update `validateRegex` to flag semicolons in regex as invalid to prevent server errors.
> 
>   - **Behavior**:
>     - Update `validateRegex` in `validate.ts` to return `false` if regex contains semicolons, preventing server errors.
>   - **Validation**:
>     - Semicolons in regex patterns are now flagged as invalid to avoid breaking `aw-server`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ActivityWatch%2Faw-webui&utm_source=github&utm_medium=referral)<sup> for 93f44842378ddeac9addb7f3d3949644649427cd. You can [customize](https://app.ellipsis.dev/ActivityWatch/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/util/validate.ts
+++ b/src/util/validate.ts
@@ -2,6 +2,10 @@ export function validateRegex(re: string) {
   // validates if pattern is a valid regex
   // returns true if regex is valid
   if (re === '') return false;
+ 
+  // semicolons are valid regex, but they break aw-server
+  if (/[;]/.test(re)) return false;
+
   try {
     new RegExp(re);
   } catch (e) {
--- a/src/util/validate.ts
+++ b/src/util/validate.ts
@@ -4,7 +4,7 @@ export function validateRegex(re: string) {
   if (re === '') return false;
  
   // semicolons are valid regex, but they break aw-server
-  if (/[;]/.test(re)) return false;
+  if (re.includes(';')) return false;
 
   try {
     new RegExp(re);


ErikBjare:
@TimeToBuildBob Investigate this, shouldn't it get escaped in the query and be supported? Does it break aw-server-python or aw-server-rust or both?

TimeToBuildBob:
Investigated both servers.

**Short answer**: semicolons *should* be supported and don't need to be blocked. Both servers handle them correctly when the regex is embedded as a string literal in the AQL query (which is what aw-webui does via `JSON.stringify`).

**aw-server-python**: Was broken — the old query2.py split statements on *all* semicolons without tracking string context. Fixed in ActivityWatch/aw-core#131 (merged 2026-02-27). The `_split_query_statements()` function now tracks quote state and skips semicolons inside string literals.

**aw-server-rust**: Never broken. The plex-based lexer in `aw-query/src/lexer.rs` tokenizes double-quoted strings first (`"([^"]|(\\"))*"` regex → `Token::String`), so a semicolon inside `"abc;def"` is consumed as part of the String token and the `Token::Semi` pattern never fires for it.

**How aw-webui builds queries**: `queries.ts` serializes categories via `JSON.stringify(params.categories)`, which wraps regex values in double quotes. So a regex `abc;def` appears in the AQL query as `"abc;def"` — safely inside a quoted string. No escaping is needed on top of that.

**Recommendation**: This PR isn't needed. The original bug (forum post) was in an older aw-server-python before the aw-core fix. The right fix was aw-core#131, which is already merged. Blocking semicolons in the UI unnecessarily restricts valid regex patterns (e.g. alternation with `;` characters in paths on some systems).

Could close this PR as not needed, or the contributor could be asked to add a test to aw-core instead verifying that semicolons in categorization rules work end-to-end.

TimeToBuildBob:
**Investigation complete — root cause found, fix in #918.**

**Short answer**: yes, semicolons should be supported. Both servers handle them correctly — the bug is in the webui's client-side query splitter, not in aw-server-python or aw-server-rust.

**What breaks**: `querystr_to_array()` in `queries.ts` naively does `.split(';')` on the whole query string. Category rules with semicolons in their regex get embedded as JSON string literals:

```
events = categorize(events, [["Work", {"type":"regex","regex":"foo;bar"}]]);
```

A plain `.split(';')` shreds this at the `;` inside the JSON string before the query ever reaches either server.

**Both servers are fine**: I checked `aw-query/src/lexer.rs` — the Rust lexer's string-literal rule (`"([^\"]|(\\\"))*"`) captures the full quoted string including any `;` inside it, so semicolons in string literals never become `Token::Semi`. The Python parser works the same way. Neither server sees the malformed query because the client breaks it first.

**Fix**: PR #918 replaces the naive split with a string-context-aware parser that only treats `;` as a statement separator when it's outside a double-quoted string literal. 4 unit tests added including the regression case.

PR #788, opened 2026-03-19
datepicker improvements

nerumo:
- support start of week
- add next/previous day buttons
- 
image


0xbrayo:
implemented in #839, close PR

PR #795, opened 2026-04-02
[IMP] add server settings for CORS

RaoufGhrissi:
Dependent on: https://github.com/ActivityWatch/aw-server-rust/pull/581


RaoufGhrissi:
I will apply the same changes to aw-server once validated on aw-server-rust

PR #455, opened 2023-06-22
feat: added improved categorization helper that includes bigrams if commonly present

ErikBjare:
Came up with this while thinking about the bug in https://github.com/ActivityWatch/activitywatch/issues/905, with the help of ChatGPT: https://chat.openai.com/share/0c7588ee-b13b-41e3-b169-c1e587fcbb72

PR #803, opened 2026-04-14
feat(categorization): extract findCommonPhrases util with bigram support and tests

TimeToBuildBob:
Takes over #455 as requested by @ErikBjare.

## Changes

- Extracts the bigram-based phrase detection into `src/util/categorization.ts` for testability
- Uses `Map<string, WordEntry>` (consistent with existing `CategoryBuilder.vue` code)
- Also filters bigram components against `ignored_words` and length ≤ 2 (the original only filtered single words)
- Removes debug `console.log` statements from the original PR
- Full TypeScript types

## Tests

10 unit tests covering:
- Empty input
- Duration accumulation
- Bigram promotion when it dominates both words (>50% threshold)
- No promotion when a word appears in many different contexts
- `ignored_words` filtering (single words and bigram components)
- Short-word filtering (length ≤ 2)
- Various separator characters (`-`, `,`, `:`, `(`, `)`)
- Multi-event accumulation

```
PASS node test/unit/categorization.test.node.ts
  ✓ returns empty map for empty events
  ✓ counts single words by duration
  ✓ promotes bigram when it dominates both constituent words
  ✓ does not promote bigram when one word appears independently too often
  ✓ filters out words with length <= 2
  ✓ filters out ignored words
  ✓ ignored words are not used as bigram components
  ✓ handles titles split by various separator characters
  ✓ accumulated duration across multiple events
  ✓ returns Map with word entries containing events list
Tests: 10 passed, 10 total
```

### DIFF
--- a/src/util/categorization.ts
+++ b/src/util/categorization.ts
@@ -64,6 +64,15 @@ export function findCommonPhrases(
   }
 
   // Step 3: Promote bigrams that dominate both constituent words (>50% threshold)
+  // Snapshot original durations before the loop so that promoting one bigram
+  // (which reduces its constituent words' durations) does not corrupt the
+  // threshold check for a later bigram that shares a common word.  Without this,
+  // a trigram such as "Alpha Beta Gamma" causes Beta.duration to reach 0 after
+  // "Alpha Beta" is promoted, making 10/0 = Infinity pass the check for "Beta Gamma"
+  // even though only 10/110 ≈ 9% of Beta's original time was spent in that bigram.
+  const originalDurations = new Map<string, number>(
+    Array.from(words.entries(), ([w, e]) => [w, e.duration])
+  );
   for (const [bigram, bigramEntry] of bigrams) {
     const spaceIdx = bigram.indexOf(' ');
     const word1 = bigram.slice(0, spaceIdx);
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Fixed the P1 bug flagged by Greptile in 99b073f.

**Root cause**: In Step 3 the promotion loop mutated `w1Entry.duration` / `w2Entry.duration` in-place. When two bigrams share a middle word (trigram case), the first promotion reduces that word's duration to 0. The next check then sees `10 / 0 = Infinity > 0.5`, so a weak bigram that should not qualify gets incorrectly promoted.

**Fix**: snapshot all word durations into `originalDurations` before the loop and use those snapshots for threshold comparisons. Mutations to `entry.duration` still happen (so individual word entries reflect the reduced "leftover" time), but they no longer corrupt subsequent threshold checks.

**Regression test added**: `'trigram: does not double-promote when middle word duration is consumed'` — sets up `makeEvent('Alpha Beta', 100)` + `makeEvent('Alpha Beta Gamma', 10)` where `"Beta Gamma"` (10/110 ≈ 9% of Beta) must not be promoted. The test fails on the unfixed code and passes after the fix.

TimeToBuildBob:
Fixed lint failure in 4a853eb: replaced the two `!` non-null assertions on `originalDurations.get()` with explicit `undefined` checks in the loop guard. Logically they were always defined (populated from the same `words` map), but ESLint correctly flagged the unsafe assertions.

TimeToBuildBob:
Fixed the remaining lint warning in 224a5a1: broke the over-length guard clause onto two lines to satisfy prettier's line-length rule (the previous commit fixed the logic but the resulting line was 100 chars, over prettier's 80-char limit).

ErikBjare:
@greptileai review

PR #809, opened 2026-04-23
fix(query-options): prefer known hosts over unknown defaults

TimeToBuildBob:
## Summary
- prefer known hosts over `unknown` in the shared QueryOptions hostname chooser
- keep `unknown` available as a fallback option when it is the only host
- add a regression test for the hostname ordering helper

## Root cause
`QueryOptions` defaulted to `hostnameChoices[0]`, and `bucketsStore.hosts` can legitimately put `unknown` first when a special bucket like `aw-stopwatch` was updated more recently than the real window/AFK buckets.

That meant Category Builder could start with `queryOptions.hostname = "unknown"` already set, which bypassed its existing "if hostname is blank, pick a non-unknown host" fallback and led to synthesized queries like `aw-watcher-window_unknown`.

## Testing
- `npm test -- --runInBand test/unit/hostnames.test.node.ts test/unit/queries.test.node.js`

Refs ActivityWatch/activitywatch#1214

### DIFF
--- a/src/components/QueryOptions.vue
+++ b/src/components/QueryOptions.vue
@@ -17,6 +17,7 @@ div
 import Vue from 'vue';
 import moment from 'moment';
 import { useBucketsStore } from '~/stores/buckets';
+import { preferKnownHostnames } from '~/util/hostnames';
 
 export default Vue.extend({
   name: 'QueryOptions',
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
All checks are green and the diff is the targeted hostname-ordering fix with regression coverage, so this is ready to merge. I attempted to merge it from project monitoring just now, but the TimeToBuildBob credentials on this runner do not have MergePullRequest permission for ActivityWatch/aw-webui, so a maintainer merge is still needed.

0xbrayo:
One of the most annoying bugs fixed. Good job @TimeToBuildBob

PR #826, opened 2026-05-20
fix: remove deprecated keyboard option from vis-timeline

fanxing11:
vis-timeline v7 dropped the `keyboard` configuration option (it was part of the original vis.js library). Passing it triggers a console warning:

```
Unknown option detected: "keyboard". Did you mean "end"?
```

This shows up as an error in the Raw Data view — reported in ActivityWatch/aw-server-rust#603.

The fix just removes the `keyboard` block from `VisTimeline.vue` options. `horizontalScroll` is kept since it's still a valid vis-timeline v7 option.

### DIFF
--- a/src/visualizations/VisTimeline.vue
+++ b/src/visualizations/VisTimeline.vue
@@ -105,12 +105,8 @@ export default {
           overflowMethod: 'flip',
           delay: 0,
         },
-        // Keyboard & scroll navigation (see #629)
+        // Horizontal scroll navigation (see #629)
         horizontalScroll: true, // horizontal scroll/swipe pans the timeline
-        keyboard: {
-          enabled: true,
-          speed: { x: 10, y: 0, zoom: 0.02 },
-        },
       },
       editingEvent: null,
       editingEventBucket: null,


ErikBjare:
Had trouble finding what happened to it, but Google AI said:

```
In vis-timeline v7, the legacy global configuration option for keyboard (which previously allowed you to pass keys or toggle default keyboard event listeners natively via options: { keyboard: ... }) was completely removed.
The library decoupled global keyboard event handlers to give you direct control over the DOM, minimize accessibility clashes, and reduce bundle weight.
To restore your keyboard shortcuts (like panning or zooming with the arrow keys), you must now map standard JavaScript listeners to the vis-timeline instance API manually.
## How to Restore Keyboard Navigation in v7
You can catch keyboard events on your container and use timeline.getWindow() and timeline.setWindow() to move or scale the view.

// 1. Initialize your timeline normallyconst container = document.getElementById('visualization');const timeline = new vis.Timeline(container, items, options);
// Ensure the container can receive focus for keyboard events
container.setAttribute('tabindex', '0'); 
// 2. Add custom keyboard listeners
container.addEventListener('keydown', (event) => {
  const range = timeline.getWindow();
  const interval = range.end - range.start;
  const percentage = 0.1; // Amount to pan/zoom per keypress

  switch (event.key) {
    case 'ArrowLeft': // Pan Left
      timeline.setWindow(range.start.valueOf() - interval * percentage, range.end.valueOf() - interval * percentage);
      event.preventDefault();
      break;
    ...
});
```

Will merge this, I guess proper restoration of the feature needs a fix like the suggested, but that's a separate PR.

ErikBjare:
Thank you for your contribution @fanxing11!

PR #835, opened 2026-05-21
fix(query-explorer): solid Rename/Delete buttons for saved queries

TimeToBuildBob:
Follow-up to #829 based on @0xbrayo's review comment (left just after merge):

> Use a solid color for the rename button, same color used in the expand list button on the same page. Use the bin button same as in the raw data section.

**Changes:**
- **Rename**: `btn-outline-secondary` → `btn-secondary` (solid secondary, matching the "Expand list" button style used in the event list on the same page)
- **Delete**: `btn-outline-danger` → `btn-danger` with `icon(name="trash")` (matches the trash-icon+solid-danger pattern used in Buckets and activity views)

No behaviour changes — purely visual/style alignment.

### DIFF
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -14,6 +14,7 @@ module.exports = {
         '^.+\\.ts$': 'ts-jest',
         '^.+\\.vue$': '@vue/vue2-jest',
       },
+      transformIgnorePatterns: ['/node_modules/(?!(vue-awesome)/)'],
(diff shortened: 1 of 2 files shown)


ErikBjare:
Not sure I agree on the solid color stuff @0xbrayo (they are secondary/non-primary buttons after all - draws attention they don't deserve when solid imo), but I'll merge it and it can be redone later.

PR #131, opened 2019-03-08
Add brave and opera to browser appnames

johan-bjareholt:
https://forum.activitywatch.net/t/web-watcher-doesnt-work-with-brave

Does not work without https://github.com/ActivityWatch/aw-watcher-web/pull/37

PR #875, opened 2026-06-17
fix(query): configurable flood pulsetime to fix missing time with high poll intervals

### LINKED ISSUE #1177: Fundamental flaw in duration calculation
- [x] I am on the [latest](https://github.com/ActivityWatch/activitywatch/releases/latest) ActivityWatch version.
 - [x] I have searched the issues of this repo and believe that this is not a duplicate.

- **OS name and version**: Windows 11
- **ActivityWatch version**: Version:  v0.13.2
## Describe the bug

I noticed that there are gabs between window watcher events. These are small, but accumulate over a day and are dependent on polling size.

The window watcher reports times from start of occurrence and calculates a duration until the next event -1* polling duration (1s gab with 1s polling time, 5s gab with 5s polling time) which is wrong and leads to gabs:

E.g. (polling time 5s)

App1: 0.0 (reported duration 5s, actual duration 10s)
App2: 0.10 (reported duration 10s, actual duration 15s)
App3: 0.25 (reported duration 0s, actual duration 5s)
App4: 0.30

-> sum of all duration (15s), actual time (30s).

This is very noticeable if polling time is set to high values. But even at 1s this will lead to problems, esp. with programs which often switch title. E.g.:

App1: 0.0 (reported duration 0s, actual duration 1s)
App2: 0.1 (reported duration 3s, actual duration 4s)
App3: 0.5

-> sum (3s), actual time (5s)

When using a stopwatch, and high polling time, you can see very clearly both are diverging.
All of this started when I tracked my daily work time and was confused why the numbers would not add up. When I work 8h it only reported 7.5h, etc.

After investigation and looking into the exported `.json` files, I found the wrongly calculated duration.

## To Reproduce

1. Change polling time to 30s
2. Use two apps >30s
3. Repeat this a few times
4. Check timeline
5. Export bucket and check calculated times

## Expected behavior

An application duration is until the next application, inclusive of the timestamp of the new application. There are no gabs.

## Documentation

Image

1s gab:

Image

1s gab:

Image

Polling time 5s (duration of explorer is 5s not 0s, others are off as well):

Image

Polling time 1s:

Image

## Additional context

Probably related issues:
- #1136  (many Window switches with a second loss per switch can result in the inaccuracy)
- #1093 (unsure)

behdadmansouri:
## Summary

Fixes ActivityWatch/activitywatch#1177

Users who raise the watcher polling interval above 5s lose trackable time because the default 5s `pulsetime` in `flood()` no longer fills the inter-poll gaps. A user working 8h could see only 7.5h in the dashboard.

**Changes:**
- `queries.ts`: Add optional `floodPulsetime` field to `BaseQueryParams` (default `5`); pass it as an explicit second argument to every `flood()` call inside `canonicalEvents()` and `browserEvents()`
- `stores/settings.ts`: Add `floodPulsetime: number` (default `5`) — persisted like all other settings
- `stores/activity.ts`: Include `floodPulsetime: useSettingsStore().floodPulsetime` in both `fullDesktopQuery` and `multideviceQuery` param objects
- `views/settings/DeveloperSettings.vue`: Add a number input for "Flood pulsetime (seconds)" with descriptive help text explaining the `polling_interval + 1` rule

## Test plan

- Default behaviour unchanged: `floodPulsetime` defaults to `5`, all existing queries behave identically
- User with 30s polling sets "Flood pulsetime" to `31` in Developer Settings → gaps are filled, reported time matches clock time
- Setting persists across reloads

> **Note:** This pairs with [ActivityWatch/aw-core#139](https://github.com/ActivityWatch/aw-core/pull/139) which exposes `pulsetime` as a parameter of the `flood()` query function. This web UI change works independently (the query language already forwards extra args correctly).

> 🤖 Implemented with [Claude Code](https://claude.ai/code) — AI assistance disclosed per contribution guidelines.

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -33,6 +33,14 @@ interface BaseQueryParams {
   bid_browsers?: string[];
   bid_stopwatch?: string;
   return_variable_suffix?: string;
+  /**
+   * Seconds used as the pulsetime argument to flood().
+   * Must be >= your watcher's polling interval to avoid gaps in the timeline.
+   * Default: 5 (matches the default 1s poll interval with headroom).
+   * Increase this if you use a higher polling interval (e.g. set to 31 for 30s polling).
+   * See: https://github.com/ActivityWatch/activitywatch/issues/1177
+   */
+  floodPulsetime?: number;
 }
 
 interface DesktopQueryParams extends BaseQueryParams {
@@ -125,14 +133,16 @@ export function canonicalEvents(params: DesktopQueryParams | AndroidQueryParams)
   // For simplicity, we assume that bid_window and bid_android are exchangeable (note however it needs special treatment)
   const bid_window = isDesktopParams(params) ? params.bid_window : params.bid_android;
 
+  const pulsetime = params.floodPulsetime ?? 5;
+
   return [
     // Fetch window/app events
-    `events = flood(${queryBucket(bid_window)});`,
+    `events = flood(${queryBucket(bid_window)}, ${pulsetime});`,
     // On Android, merge events to avoid overload of events
     isAndroidParams(params) ? 'events = merge_events_by_keys(events, ["app"]);' : '',
     // Fetch not-afk events
     isDesktopParams(params)
-      ? `not_afk = flood(${queryBucket(params.bid_afk)});
+      ? `not_afk = flood(${queryBucket(params.bid_afk)}, ${pulsetime});
          not_afk = filter_keyvals(not_afk, "status", ["not-afk"]);` +
         (always_active_pattern_str
           ? `not_treat_as_afk = filter_keyvals_regex(events, "app", "${always_active_pattern_str}");
@@ -279,13 +289,14 @@ export const browser_appname_regex: Record<string, string> = {
 
 // Returns a list of active browser events (where the browser was the active window) from all browser buckets
 function browserEvents(params: DesktopQueryParams): string {
+  const pulsetime = params.floodPulsetime ?? 5;
   let code = `
     browser_events = [];
   `;
 
   _.each(browsersWithBuckets(params.bid_browsers), ([browserName, bucketId]) => {
     const browser_appnames_str = JSON.stringify(browser_appnames[browserName]);
-    code += `events_${browserName} = flood(query_bucket("${bucketId}"));
+    code += `events_${browserName} = flood(query_bucket("${bucketId}"), ${pulsetime});
        window_${browserName} = filter_keyvals(events, "app", ${browser_appnames_str});`;
 
     // Add regex-based matching to cover case/spacing/versioning variants (e.g., Firefox.exe, firefox-esr-esr140)
--- a/src/stores/activity.ts
+++ b/src/stores/activity.ts
@@ -342,6 +342,7 @@ export const useActivityStore = defineStore('activity', {
         filter_categories,
         host_params: {},
         always_active_pattern,
+        floodPulsetime: useSettingsStore().floodPulsetime,
       });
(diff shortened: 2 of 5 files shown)


0xbrayo:
the python server won't be supported in the future. Should make a PR to the rust server. Keep the python PR too for compatibility.

TimeToBuildBob:
**Review from @TimeToBuildBob**

Cross-referencing with my analysis in ActivityWatch/activitywatch#1177 — this implementation looks clean and addresses the core problem correctly.

### What's good

- `floodPulsetime` defaults to `5`, so **all existing behavior is unchanged** for the vast majority of users
- The JSDoc on `BaseQueryParams` and the description in `DeveloperSettings.vue` both include the `polling_interval + 1` rule, which is the key user-facing hint
- `v-model.number` in the Vue input avoids the string-vs-number pitfall
- The setting propagates through both `fullDesktopQuery` and `multideviceQuery` (the two query paths that call `flood()`)
- The `Buckets.vue` import success/error alert improvement is a nice bonus (and fixes the `dismissable` → `dismissible` typo)

### UX trade-off to note

My suggestion in #1177 was to auto-derive pulsetime from the watcher's polling config so users don't have to manually remember to update a setting when they change their polling interval. The manual approach here is simpler to implement but creates a potential UX gap: a user who sets polling to 30s will still see missing time until they discover the Developer Settings entry.

That said, the auto-derive approach requires reading per-bucket poll intervals (not trivially available in the query layer), and may be worth a follow-up issue rather than blocking this fix. The current PR is already a significant improvement for anyone experiencing the gap.

### CI note

The fork's CI didn't run, so there's no automated green check here. Given the diff is UI-only (TypeScript + Vue), quick local verification: build passes, settings persist, pulsetime flows into the query string correctly.

### Recommendation

Ready to merge as a pragmatic fix for the missing-time problem. If auto-detection is wanted later, ActivityWatch/aw-core#139 (expose pulsetime as a parameter) is the upstream piece that would enable it.

ErikBjare:
I think the root cause was fixed in https://github.com/ActivityWatch/aw-watcher-window/pull/131?

TimeToBuildBob:
Yes. ActivityWatch/aw-watcher-window#131 (merged 2026-07-28) is the ingest-time fix for ActivityWatch/activitywatch#1177, which is now closed.

Two different pulsetimes were in play:

- **Heartbeat pulsetime** (watcher → server): consecutive same-window samples merge at ingest. That was the actual missing-time bug — at `poll_time=5s` the old `poll_time+1` margin lost ~12% of session time to scheduling jitter. #131 changed it to `max(poll_time*1.5, poll_time+1)`.
- **Query flood pulsetime** (this PR): fills gaps between already-stored events at analysis time. Default is still 5s. That's a workaround for data that already has gaps, not the cause of the 8h→7.5h undercount.

This PR is also the wrong layer going forward: aw-server-rust's `flood()` still takes exactly one argument and hardcodes 5s (`validate::args_length(&args, 1)`). Passing `flood(events, 31)` from the web UI would error on the rust server (as @0xbrayo flagged).

I'd close this as superseded by #131. The remaining query-flood default of 5s is a smaller, separate issue (inter-app gaps at very high poll_time). If that should scale, it belongs in rust `flood()` plus auto-derive from watcher poll interval — not a Developer Settings knob.

Unrelated `Buckets.vue` import-success UI in this diff is still open as ActivityWatch/activitywatch#394; worth a dedicated PR if anyone still wants it.

Thanks @behdadmansouri for the work — the diagnosis of the 5s flood default was right; the durable fix just landed one layer down, in the watcher.

PR #884, opened 2026-06-22
fix(tests): skip cross-locale assertions when Node lacks full-icu

### LINKED ISSUE #871: FAIL jsdom test/unit/time.test.js
Are these test run somewhere? I try to build the latest master at [COPR](https://download.copr.fedorainfracloud.org/results/wojnilowicz/playground/fedora-44-x86_64/10606267-nodejs-aw-webui/builder-live.log.gz) and get an error as below.
`FAIL jsdom test/unit/time.test.js
  ● locale-aware time labels › should format weekday labels using the supplied locale

    expect(received).not.toBe(expected) // Object.is equality

    Expected: not "Mon"

      42 |       new Intl.DateTimeFormat('sv-SE', { weekday: 'short' }).format(monday)
      43 |     );
    > 44 |     expect(format_weekday_short(monday, 'sv-SE')).not.toBe(format_weekday_short(monday, 'en-US'));
         |                                                       ^
      45 |   });
      46 |
      47 |   test('should format month-day labels without hardcoded english ordinals', () => {

      at Object.toBe (test/unit/time.test.js:44:55)`

fanxing11:
Fixes #871.

When Node.js is built without full ICU data (the default on Debian/Fedora/Arch distro packages), `Intl.DateTimeFormat('sv-SE', …)` silently falls back to the default locale. The output then matches `'en-US'`, so the `.not.toBe` / `.not.toEqual` assertions in `test/unit/time.test.js` fail when building from source. CI is green because GitHub Actions runners ship full-icu.

Picked option A from @TimeToBuildBob's analysis in the issue — gate the two cross-locale assertions on a runtime ICU check, no new dependency. Positive assertions still run, so behavior on full-icu builds is unchanged.

Tested locally (Node 22 with full-icu): 24/24 tests pass.

```js
function hasLocale(locale) {
  try { return new Intl.DateTimeFormat(locale).resolvedOptions().locale === locale; }
  catch (_) { return false; }
}
```

### DIFF
--- a/test/unit/time.test.js
+++ b/test/unit/time.test.js
@@ -31,6 +31,20 @@ describe('seconds_to_duration', () => {
   });
 });
 
+// Node builds without full ICU data silently fall back to the default locale
+// for any non-English locale, which breaks cross-locale assertions. This is
+// common in distro-packaged Node (Debian, Fedora, Arch). Detect it so we can
+// skip the cross-locale comparisons in those environments. See issue #871.
+function hasLocale(locale) {
+  try {
+    return new Intl.DateTimeFormat(locale).resolvedOptions().locale === locale;
+  } catch (_) {
+    return false;
+  }
+}
+
+const HAS_SV_SE = hasLocale('sv-SE');
+
 describe('locale-aware time labels', () => {
   test('should format weekday labels using the supplied locale', () => {
     const monday = new Date(2026, 4, 18, 12);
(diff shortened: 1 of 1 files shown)


TimeToBuildBob (approved):
Looks good! Clean implementation of Option A from the scouting analysis in #871.

Key verification:
- **`hasLocale` semantics**: `Intl.DateTimeFormat(locale).resolvedOptions().locale === locale` correctly detects full-ICU by checking if the runtime actually resolved the requested locale (partial-ICU silently falls back to default, breaking the strict equality)
- **Scope**: Only the two cross-locale assertions are gated; all positive/locale-invariant assertions run unconditionally — so test coverage on CI (full-ICU) is completely unchanged
- **No new dependencies**: exactly as specified

Tested locally (Node 22 with full-icu): 24/24 ✅

Thanks @fanxing11 for the contribution!

PR #890, opened 2026-07-01
fix(buckets): surface real server error message on import failure

TimeToBuildBob:
## Summary

Bucket import in the web UI always showed the generic message
`Import failed, see aw-server logs for more info` on any failure, even
though aw-server-rust has returned the actual error in the JSON
response body (`{"message": "..."}`) since `HttpErrorJson` was
introduced — the code just never read it (there's a `// TODO: Make
aw-server report error message so it can be shown in the web-ui` left
in `Buckets.vue`, but the server side of that TODO was already done).

- Read `err.response.data.message` when present and show it, falling
  back to the generic message otherwise. Same pattern already used in
  `Search.vue`, `QueryExplorer.vue`, `Graph.vue`, `Report.vue`,
  `Alerts.vue`, and `ErrorBoundary.vue`.
- Also show a success alert ("Import completed successfully!") after a
  successful import — previously nothing was displayed on success at
  all, which was part of the original feature request too.

Fixes #394

## Test plan

- [x] `npx vue-cli-service lint` on the changed file — no errors
- [x] `npm run build` — full production build completes with no errors
- [ ] Manual click-through (import a valid file, an invalid file, and
      re-import an existing bucket) — not run here; happy to do so if a
      maintainer wants a screenshot/GIF before merge

### DIFF
--- a/src/views/Buckets.vue
+++ b/src/views/Buckets.vue
@@ -138,6 +138,8 @@ div
     b-card(header="Import buckets")
       b-alert(v-if="import_error" show variant="danger" dismissible)
         | {{ import_error }}
+      b-alert(v-if="import_success" show variant="success" dismissible)
+        | Import completed successfully!
       b-form-file(v-model="import_file"
                   placeholder="Choose or drop a file here..."
                   drop-placeholder="Drop file here...")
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
All 9 checks are now green including the last one (Test on aw-server v0.12.3b18). Greptile 5/5. Ready for human review whenever someone has a moment.

Fixes #394.

PR #898, opened 2026-07-04
fix: match Helium with chrome web watcher bucket

vedntp:
## Summary

- Match `Helium` for Chrome web watcher buckets when no dedicated Helium bucket exists.
- Avoid matching `Helium` through the Chrome bucket when an `aw-watcher-web-helium_*` bucket is also present.
- Add regression tests for both bucket configurations.

## Why

Helium can install and run the Chrome Web Store ActivityWatch extension, which reports URL events into an `aw-watcher-web-chrome_*` bucket. In that setup, the active window app name is `Helium`, so the Chrome bucket filter would otherwise drop those browser events even though the data exists.

The PR now also guards the edge case where both a Chrome-extension bucket and a dedicated Helium bucket are present, so the same Helium browsing activity is not claimed by both browser paths.

## Test

```sh
npm test -- --selectProjects node --runTestsByPath test/unit/queries.test.node.ts
npx prettier --check src/queries.ts test/unit/queries.test.node.ts
```

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -263,7 +263,8 @@ function browsersWithBuckets(browserbuckets: string[]): [string, string][] {
 // The full set of historical app names these patterns replace is documented in the unit tests.
 // See: test/unit/queries.test.node.ts, https://github.com/ActivityWatch/aw-webui/issues/749
 export const browser_appname_regex: Record<string, string> = {
-  chrome: '(?i)^(google[-_ ]?chrome|chrome|chromium)',
+  // Helium can use the Chrome Web Store extension, which emits aw-watcher-web-chrome buckets.
+  chrome: '(?i)^(google[-_ ]?chrome|chrome|chromium|helium)',
   firefox: '(?i)(firefox|librewolf|waterfox|nightly)',
   opera: '(?i)(opera)',
   brave: '(?i)(brave)',
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -245,11 +245,21 @@ const browser_appnames: Record<string, string[]> = {
 };
 
 // Returns a list of (browserName, bucketId) pairs for found browser buckets
+function browserBucketMatches(bucketId: string, browserName: string): boolean {
+  return _.includes(bucketId, browserName);
+}
+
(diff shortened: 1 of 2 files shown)


TimeToBuildBob (approved):
Reviewed the diff carefully. This is a clean, well-designed fix.

**What it does**: When Helium is running with the Chrome Web Store extension (reports to `aw-watcher-web-chrome_*`), the chrome query now includes `helium` in the app-name regex — but only when there's no dedicated `aw-watcher-web-helium_*` bucket, preventing double-counting when a proper Helium bucket exists.

**Code quality**: Good factoring — `browserBucketMatches`/`browserBucketExists` helpers keep the logic readable, and `browserAppnameRegex` cleanly handles the conditional case. Tests cover both configurations (with and without a dedicated Helium bucket), which is exactly the right coverage.

**No concerns**: Backward compatible — no behavior change when Helium is absent. The conditional regex routing is straightforward. Greptile also passed.

LGTM ✓

TimeToBuildBob:
Noting this was closed rather than merged — the Helium-in-chrome-bucket case is not in `master`, so flagging it here so it doesn't get lost:

- Current `src/queries.ts` chrome bucket pattern is `(?i)^(google[-_ ]?chrome|chrome|chromium|arc(\.exe)?$|dia(\.exe)?$)` — no `helium`. `helium` exists only as its own key (`net.imput.helium` / `aw-watcher-web-helium_*`).
- #928 added Arc/Dia to the chrome pattern and #744 added the standalone Helium bucket, but neither covers the setup this PR fixes: the Chrome build of the extension running inside Helium, with events landing in `aw-watcher-web-chrome_*` while the active window app is `Helium`.
- #949's `chromeAppnameRegex(excludeForks)` / `CHROME_FORK_ALTS` is the generalized form of this PR's design (same conditional-exclude idea, arc/dia instead of helium). It is green and awaiting merge; adding `helium` to that map reproduces this fix — double-count guard included — without a second code path.

The conditional-regex design here was the right call; thanks for it. I've tracked folding Helium into the #949 mechanism as a follow-up so it lands with that PR.

PR #903, opened 2026-07-12
fix(activity): detect Android device at runtime for view selection

TimeToBuildBob:
## Problem

`VUE_APP_ON_ANDROID` is set at **build time** and only covers the case where aw-webui is embedded inside the Android app itself. When a desktop browser views data from an Android device, the flag is always `false`, so the default views (`desktopViews`) are used — showing "Top Window Titles" and other desktop-only elements instead of the Android summary view.

This was noted in a `FIXME` comment in `views.ts`:
```
// FIXME: Decide depending on what kind of device is being viewed, not from which device it is being viewed from.
```

The symptom: browsing to an Android device in aw-webui from a desktop shows "Top Window Titles" and a desktop device icon instead of the Android summary.

## Fix

Replace `...mapState(useViewsStore, ['views'])` in `Activity.vue` with an explicit computed property that checks `bucketsStore.available(host).android` at runtime. If the host has Android buckets but no window buckets (i.e. it is an Android-only device), return `androidViews` rather than the stored default.

This also exports `androidViews` from `views.ts` so it can be imported in `Activity.vue`.

## Behaviour

| Situation | Before | After |
|-----------|--------|-------|
| Desktop browser viewing Android device data | Desktop views (broken) | Android summary view ✓ |
| Desktop browser viewing desktop data | Desktop views ✓ | Desktop views ✓ |
| Android app (VUE_APP_ON_ANDROID=true) | Android views ✓ | Android views ✓ |
| Mixed host (both android + window buckets) | Desktop views | Desktop views (conservative) |

## Note on hostname spaces

This also indirectly helps with OEM hostnames that contain spaces (e.g. `"Poco F8 Ultra"` from `Build.MODEL`). The `bucketsAndroid` detection in `buckets.ts` already handles the hostname comparison correctly — the primary issue was the view selection using build-time env rather than runtime bucket detection.

### DIFF
--- a/src/views/activity/Activity.vue
+++ b/src/views/activity/Activity.vue
@@ -229,8 +229,7 @@ import 'vue-awesome/icons/ellipsis-v';
 import { useSettingsStore } from '~/stores/settings';
 import { useCategoryStore } from '~/stores/categories';
 import { useActivityStore, QueryOptions } from '~/stores/activity';
-import { useViewsStore, androidViews } from '~/stores/views';
-import { useBucketsStore } from '~/stores/buckets';
+import { useViewsStore } from '~/stores/views';
 
 export default {
   name: 'Activity',
@@ -255,7 +254,6 @@ export default {
   data: function () {
     return {
       activityStore: useActivityStore(),
-      bucketsStore: useBucketsStore(),
       categoryStore: useCategoryStore(),
       viewsStore: useViewsStore(),
       settingsStore: useSettingsStore(),
@@ -278,15 +276,7 @@ export default {
   },
   computed: {
     views(): import('~/stores/views').View[] {
-      // Use Android views when the current host is an Android device.
-      // VUE_APP_ON_ANDROID is a build-time flag that only applies when the webui
-      // is embedded in the Android app itself. When browsing Android device data
-      // from a desktop browser we need a runtime check instead.
-      const available = this.bucketsStore.available(this.host);
-      if (available.android && !available.window) {
-        return androidViews;
-      }
-      return this.viewsStore.views;
+      return this.viewsStore.viewsForHost(this.host);
     },
     ...mapState(useSettingsStore, ['devmode']),
     ...mapState(useSettingsStore, ['always_active_pattern']),
--- a/src/views/activity/Activity.vue
+++ b/src/views/activity/Activity.vue
@@ -229,7 +229,8 @@ import 'vue-awesome/icons/ellipsis-v';
 import { useSettingsStore } from '~/stores/settings';
 import { useCategoryStore } from '~/stores/categories';
 import { useActivityStore, QueryOptions } from '~/stores/activity';
-import { useViewsStore } from '~/stores/views';
+import { useViewsStore, androidViews } from '~/stores/views';
+import { useBucketsStore } from '~/stores/buckets';
 
 export default {
   name: 'Activity',
@@ -254,6 +255,7 @@ export default {
   data: function () {
     return {
       activityStore: useActivityStore(),
+      bucketsStore: useBucketsStore(),
       categoryStore: useCategoryStore(),
       viewsStore: useViewsStore(),
       settingsStore: useSettingsStore(),
@@ -275,7 +277,17 @@ export default {
     };
   },
   computed: {
-    ...mapState(useViewsStore, ['views']),
+    views(): import('~/stores/views').View[] {
+      // Use Android views when the current host is an Android device.
+      // VUE_APP_ON_ANDROID is a build-time flag that only applies when the webui
+      // is embedded in the Android app itself. When browsing Android device data
+      // from a desktop browser we need a runtime check instead.
+      const available = this.bucketsStore.available(this.host);
+      if (available.android && !available.window) {
+        return androidViews;
+      }
+      return this.viewsStore.views;
+    },
     ...mapState(useSettingsStore, ['devmode']),
     ...mapState(useSettingsStore, ['always_active_pattern']),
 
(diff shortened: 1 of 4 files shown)


src/views/activity/Activity.vue
```diff
@@ -275,7 +277,17 @@ export default {
     };
   },
   computed: {
-    ...mapState(useViewsStore, ['views']),
+    views(): import('~/stores/views').View[] {
+      // Use Android views when the current host is an Android device.
+      // VUE_APP_ON_ANDROID is a build-time flag that only applies when the webui
+      // is embedded in the Android app itself. When browsing Android device data
+      // from a desktop browser we need a runtime check instead.
```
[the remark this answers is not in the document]
TimeToBuildBob: Good catch — this was a real gap. `ActivityView.vue` was mapped directly to `viewsStore.views` (the desktop views), so even when `Activity.vue` correctly showed the Android tabs the child was still resolving its elements from the stored desktop view definition.

Fixed in the latest commit (`317a7f5`): `ActivityView.vue` now runs the same runtime bucket check (`available.android && !available.window`) using `$route.params.host`, and returns `androidViews` when the host is an Android device. The stored views (and all edit/save operations) remain unaffected for desktop hosts.

TimeToBuildBob:
@greptileai review

src/views/activity/ActivityView.vue
```diff
@@ -73,10 +73,23 @@ export default {
...
-    return { editing: false };
+    return { editing: false, bucketsStore: useBucketsStore() };
   },
   computed: {
-    ...mapState(useViewsStore, ['views']),
+    views: function () {
+      // Use Android views when the current host is an Android device.
+      // Mirrors the same runtime check in Activity.vue so the child view
+      // resolves its elements from the correct view list instead of always
+      // falling back to the stored desktop views.
```
ErikBjare: What? So now users cant edit/persist the views? "stored desktop views" shouldn't exist in a clean install, the android views should be persisted as the defaults for that device or however it works (I guess views aren't per-device? thus not a separate "android" view being persisted?).

Related to https://github.com/ActivityWatch/aw-webui/pull/903/changes#r3566418786

useViewsStore() should handle it, not complicate views computed's

ErikBjare (changes requested):
See https://github.com/ActivityWatch/aw-webui/pull/903/changes#r3566433037

TimeToBuildBob:

ErikBjare:
@TimeToBuildBob No it isn't, see https://github.com/ActivityWatch/aw-webui/pull/903#issuecomment-4951444307 and https://github.com/ActivityWatch/aw-webui/pull/903#discussion_r3566433037

src/views/activity/ActivityView.vue
TimeToBuildBob: Good call. Pushed a refactor in `48f1fed` that moves the detection into the store:

- Added `viewsForHost(host: string): View[]` action on `useViewsStore` that does the `available.android && !available.window` check internally.
- Both `Activity.vue` and `ActivityView.vue` now just call `viewsStore.viewsForHost(this.host)` — no direct `androidViews` import, no duplicated bucket logic, no `bucketsStore` in component data.

On persistence: you're right that views are not per-device today, so for an Android host `viewsForHost` returns the hardcoded `androidViews` (not stored/customizable). The "Restore defaults" button remains wired to the store's `defaultViews` for the stored (desktop) views. Making views per-device so Android defaults persist separately would be a follow-up — that seems out of scope here, and the store is now the right place to put that logic when it happens.

TimeToBuildBob:
Fair point — I shouldn't have posted "CI-green and mergeable" while your `CHANGES_REQUESTED` review was still outstanding. That was premature. Apologies.

To be clear about what `48f1fed` actually changed (since it came in after your CHANGES_REQUESTED at 13:58): the detection logic *is* now in the store via `viewsForHost(host)` — both `Activity.vue` and `ActivityView.vue` just call `viewsStore.viewsForHost(this.host)` and hold no bucket logic themselves. So the "don't complicate views computed's" part of your concern is addressed.

What `48f1fed` does NOT address is the persistence gap you also raised: `viewsForHost` returns hardcoded `androidViews` for Android hosts, bypassing the stored/customizable view state. As I noted earlier, views aren't currently per-device, so there's no "Android views slot" to persist into.

Two options I see:

1. **Keep this PR scoped to view-selection-only**: Android hosts get the correct summary view (not broken desktop view), but the views aren't customizable for Android hosts. A follow-up adds per-device persistence. The broken behaviour is fixed; the missing customisation is a separate concern.

2. **Scope this PR wider**: Add per-device views to the store state before merging, so Android defaults can be persisted and edited independently.

I won't push more changes without your direction here. Which scope do you want?

ErikBjare:
@TimeToBuildBob 
idk if "per-device views" is the right concept, maybe more like "per-devicetype views"? Regardless, merging this, but get to work on the follow-up work that makes Android views editable (and works to edit them separate from desktop views on e.g. desktop with both desktop and mobile devices available), maybe something akin to the "category sets" (only a single `default` bundled) would be a "views set" with `desktop` and `mobile`/`android` defaults (falling back to respective default views if never persisted/edited etc).

TimeToBuildBob:
"Per-devicetype views" is the better framing — updated the tracking issue title accordingly.

Opened #904 to track the follow-up: per-devicetype views set, analogous to category sets, with `desktop` and `android` defaults that can be persisted and edited independently. `viewsForHost` would resolve to the right set based on bucket profile, falling back to bundled defaults if the user hasn't customised it.

PR #914, opened 2026-07-22
Add subscribe CTA to the Home Support section

TimeToBuildBob:
Draft for iteration — the in-app half of the ActivityWatch Pro subscription funnel (pairs with activitywatch.github.io#64, which adds the `/subscribe` page).

## What this does
Adds one tasteful line at the top of the existing **Support us!** section on the Home view:

> The best way to keep ActivityWatch maintained is a subscription — from $5/month, every feature stays free, cancel anytime. **[Support ActivityWatch →](https://activitywatch.net/subscribe?src=inapp-home)**

- Points at the new `activitywatch.net/subscribe` page with a **`?src=inapp-home`** tag so we can attribute clicks **at the destination** — the app itself phones home nothing and gets no analytics (privacy ethos preserved).
- **Patronage framing**: subscribing funds the project; every feature stays free and ungated.
- New i18n keys `home.subscribeIntro` / `home.subscribeCta` added to **all** locales (en, de, uk, ru, zh-CN); non-en use the English string as a placeholder for translators. `check:locales` passes.

## Notes for iteration
- Deliberately minimal + non-intrusive (a line in the existing support block, not a modal/banner). Easy to expand later to a dismissable banner or a report-view footer if you want more surfaces.
- The `?src=` tag assumes the counting-redirect instrumentation from the profitability plan; until that's live the link still works, it just isn't counted.
- Copy/placement are a starting point — iterate freely.

Context: existing in-app nudges have zero click instrumentation, so this destination-tagged approach is how we start actually measuring conversion.

### DIFF
--- a/src/i18n/locales/de.ts
+++ b/src/i18n/locales/de.ts
@@ -191,6 +191,9 @@ export default {
     spreadFollow: 'Folgen Sie uns und teilen Sie auf',
     spreadStar: 'Geben Sie uns einen Stern auf',
     supportTitle: 'Unterstützen Sie uns!',
+    subscribeIntro:
+      'The best way to keep ActivityWatch maintained is a subscription — from $5/month, every feature stays free, cancel anytime.',
+    subscribeCta: 'Support ActivityWatch →',
     support1: 'Gefällt Ihnen ActivityWatch? Helfen Sie uns, Ihnen zu helfen — spenden Sie über:',
     support2: 'Mehr Infos auf der',
     donationPage: 'Spendenseite der Website',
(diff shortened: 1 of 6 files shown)


TimeToBuildBob:

TimeToBuildBob:
Updated: now that [activitywatch.net/go/](https://activitywatch.net/go/?src=inapp-home) is live (shipped in activitywatch.github.io#65), the CTA routes through `/go/?src=inapp-home` instead of linking `/subscribe` directly. The redirect records a per-`src` `nudge_click` event in GA, then forwards to /subscribe — a direct link would only produce a pageview, so this is what actually captures the click. No app-side analytics; the counting happens at the destination.

PR #918, opened 2026-07-25
fix(queries): don't split on semicolons inside string literals

TimeToBuildBob:
## Root cause

`querystr_to_array()` naively splits on all `;` characters. Category rules get embedded as JSON string literals inside the query — for example a rule with regex `foo;bar` becomes:

```
events = categorize(events, [["Work", {"type": "regex", "regex": "foo;bar"}]]);
```

A plain `.split(';')` shreds this into two broken fragments before the query reaches aw-server, causing a parse error. **Both aw-server-python and aw-server-rust handle `;` inside quoted string literals correctly** — the bug is entirely in the client-side splitter.

This explains the forum report in #724 and why the UI warning (PR #724) was only a workaround.

### `src/queries.ts`
The new `querystr_to_array` tracks double-quoted string context (with backslash-escape support) and only treats `;` as a statement separator when it appears _outside_ a string literal.

Also exports `querystr_to_array` for direct unit testing and adds 4 tests:
- basic multi-statement split
- semicolon inside string literal (the regression case)
- escaped double-quote inside string
- whitespace/empty statement filtering

### View files (7 updated)
All view-level query splitting paths now use `querystr_to_array` instead of a naive `.split(';')`:
- `views/Alerts.vue`, `views/Search.vue`, `views/Graph.vue`, `views/Report.vue`, `views/QueryExplorer.vue` — replace `.split(';').map(s => s.trim() + ';')`
- `views/Timeline.vue`, `views/Trends.vue` — replace the longer split/trim/filter/map chain

## Investigation notes (from #724 comment thread)

Checked aw-server-rust's lexer (`aw-query/src/lexer.rs`): the string-literal rule `"([^\"]|(\\\"))*"` is matched before the bare `;` → `Token::Semi` rule, so semicolons inside quoted strings are correctly captured as part of `Token::String`. The server is not the problem.

Fixes #724.

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -2,12 +2,38 @@ import _ from 'lodash';
 
 // TODO: Sanitize string input of buckets
 
-function querystr_to_array(querystr: string): string[] {
-  return querystr
-    .split(';')
-    .map(s => s.trim())
-    .filter(s => s)
-    .map(s => s + ';');
+export function querystr_to_array(querystr: string): string[] {
+  // Split on ';' but only when not inside a double-quoted string literal.
+  // A naive .split(';') breaks category rules whose regex contains semicolons,
+  // e.g. `"regex": "foo;bar"` would be shredded before reaching the server.
+  const statements: string[] = [];
+  let current = '';
+  let inString = false;
+  let escape = false;
+
+  for (const char of querystr) {
+    if (escape) {
+      current += char;
+      escape = false;
+    } else if (char === '\\' && inString) {
+      current += char;
+      escape = true;
+    } else if (char === '"') {
+      inString = !inString;
+      current += char;
+    } else if (char === ';' && !inString) {
+      const trimmed = current.trim();
+      if (trimmed) statements.push(trimmed + ';');
+      current = '';
+    } else {
+      current += char;
+    }
+  }
+
+  const trimmed = current.trim();
+  if (trimmed) statements.push(trimmed + ';');
+
+  return statements;
 }
 
 function escape_doublequote(s: string) {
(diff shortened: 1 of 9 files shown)


TimeToBuildBob:
CI flags two lint warnings (treated as errors with `--max-warnings=0`):

1. `src/queries.ts:12` — `no-shadow`: the inner `escape` variable shadows the global `escape`. Renaming it (e.g. `isEscaped`) will fix this.
2. `test/unit/queries.test.node.ts:246` — prettier: trailing newline. `npx prettier --write test/unit/queries.test.node.ts` should clear it.

Greptile also flagged a P1 (confidence 3/5): view-level query execution paths apparently still call `split(';')` directly rather than going through the new `querystr_to_array`. So the regression case (category regex with semicolons) would still fire for those paths. That looks like a real gap worth addressing before merge.

TimeToBuildBob:
Both issues are now fixed and pushed.

**Lint (now passing):**
- Renamed `escape` → `inEscape` in `queries.ts` to avoid shadowing the JS global
- Collapsed the two-line string assignment in the test to satisfy prettier

**View-level scope (Greptile P1 addressed):**
Extended `querystr_to_array` to all seven views that were still doing a naive `.split(';')`:
- `Alerts.vue`, `Search.vue`, `Graph.vue`, `Report.vue`, `QueryExplorer.vue` — replaced `query.split(';').map(s => s.trim() + ';')`
- `Timeline.vue`, `Trends.vue` — replaced the longer 4-step split/trim/filter/map chain

CI is running on the updated branch.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:

PR #919, opened 2026-07-26
fix(activity): hide unsupported visualizations instead of showing warning

### LINKED ISSUE #904: feat(views): per-devicetype views set — make Android views editable and persistent
## Background

PR #903 (merged 2026-07-12) fixed Android device detection by selecting views at runtime based on bucket presence, but it uses hardcoded `androidViews` — there's no way for a user to customise or persist the Android view layout separately from the desktop layout.

## Problem

Currently, `useViewsStore` stores a single `views` array that isn't scoped to device type. When viewing an Android device, #903 returns the hardcoded `androidViews` instead of consulting the store, so:

- Android users can't edit their default view layout
- A desktop user who has both desktop and Android devices has no way to set different layouts for each device type
- The naming "per-device views" is slightly off — the right granularity is **per-devicetype** (all Android devices share a view set; all desktop devices share another)

## Proposed Solution

Introduce a **"views set"** concept, analogous to category sets:

- Two built-in sets: `desktop` and `android` (or `mobile`)
- Each set has its own default views that can be persisted and edited independently
- `useViewsStore.viewsForHost(host)` resolves to the appropriate set based on the host's bucket profile
- Falls back to the bundled defaults if the set has never been customised
- UI to switch/edit views is scoped to the active device-type

This would keep the runtime detection from #903 intact while adding the persistence and editability layer on top.

## Acceptance Criteria

- [ ] `useViewsStore` stores views keyed by devicetype (`desktop` / `android`)
- [ ] Viewing an Android device loads the persisted Android views (falls back to `androidViews` default)
- [ ] Users can edit and persist Android views independently of desktop views
- [ ] A desktop user with both device types can customise each independently
- [ ] No regression for the existing desktop view editing flow

TimeToBuildBob:
Closes #904.

## What this does

Replaces the per-devicetype view-set approach from #903 (which returned hardcoded `androidViews` for Android devices) with a simpler model: **one view layout for all devices, with individual visualizations hidden when they have no backing data**.

Erik's direction from #904:
> "let's stick with hiding unsupported visualizations for now. That way visualizations can also 'automatically' start showing once the requisites are there."

## Changes

**`src/components/SelectableVisualization.vue`** — add `v-if` on the root div:
```diff
-div
+div(v-if="editable || !activityStore.buckets.loaded || has_prerequisites")
```
- In **view mode**: if `has_prerequisites` is false (no backing data for this device type), the widget disappears entirely instead of showing a warning
- In **edit mode**: still shown (including the existing "missing data" warning), so users can manage their layout
- While **buckets are loading**: still shown (to prevent a flash of missing widgets)

**`src/stores/views.ts`** — simplify `viewsForHost`:
```diff
-viewsForHost(host: string): View[] {
-  const bucketsStore = useBucketsStore();
-  const available = bucketsStore.available(host);
-  if (available.android && !available.window) {
-    return androidViews;
-  }
+viewsForHost(_host: string): View[] {
   return this.views;
 },
```
The Android-specific view selection is no longer needed — the visibility guard handles it automatically.

## Behaviour

For an Android device viewed from desktop (no window buckets):
- **Summary tab**: `top_apps` visible, `top_titles` hidden, rest visible
- Visualizations reappear automatically if a watcher is later installed

For the multi-device view: no change (aggregation view uses whatever's available).

### DIFF
--- a/src/components/SelectableVisualization.vue
+++ b/src/components/SelectableVisualization.vue
@@ -1,5 +1,5 @@
 <template lang="pug">
-div(v-if="editable || !activityStore.buckets.loaded || has_prerequisites")
+div(v-if="editable || !activityStore.buckets.loaded || has_prerequisites || !settingsStore.hideUnsupportedVisualizations")
   h5
     icon.handle(name="bars" v-if="editable" style="opacity: 0.6; cursor: grab;")
     | {{ visualizations[type].title }}
@@ -131,6 +131,7 @@ import { useActivityStore } from '~/stores/activity';
 import { useCategoryStore } from '~/stores/categories';
 import { useBucketsStore } from '~/stores/buckets';
 import { useViewsStore } from '~/stores/views';
+import { useSettingsStore } from '~/stores/settings';
 
 import moment from 'moment';
 
@@ -153,6 +154,7 @@ export default {
     return {
       activityStore: useActivityStore(),
       categoryStore: useCategoryStore(),
+      settingsStore: useSettingsStore(),
 
       types: [
         'top_apps',
--- a/src/components/SelectableVisualization.vue
+++ b/src/components/SelectableVisualization.vue
@@ -1,5 +1,5 @@
 <template lang="pug">
-div
+div(v-if="editable || !activityStore.buckets.loaded || has_prerequisites")
   h5
     icon.handle(name="bars" v-if="editable" style="opacity: 0.6; cursor: grab;")
     | {{ visualizations[type].title }}
--- a/src/stores/settings.ts
+++ b/src/stores/settings.ts
@@ -66,6 +66,9 @@ interface State {
   useMultidevice: boolean;
   requestTimeout: number;
 
+  // Whether to hide visualizations that lack required data (default: off)
+  hideUnsupportedVisualizations: boolean;
+
   // Set to true if settings loaded
   _loaded: boolean;
 }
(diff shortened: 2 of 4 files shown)


ErikBjare:
@TimeToBuildBob but when should the user get notified about it? Will it still be hidden during edit? Should probably show then, right?

TimeToBuildBob:
Good questions — let me clarify both:

**Edit mode**: Yes, hidden visualizations still show in edit mode. The `v-if` condition is:
```
v-if="editable || !activityStore.buckets.loaded || has_prerequisites"
```
The `editable` term is first — so any widget that would be hidden in view mode is still visible (with its existing "missing data" warning) when the user is editing their layout. That way they can still reorganize or remove unsupported widgets.

**View mode notifications**: Currently there's no explicit notification when a widget disappears — it just silently hides once buckets finish loading. The reasoning was that this is the "positive" version of the behavior (no red warning, just a clean layout with whatever's supported). But I could see the argument for a one-time toast or an info banner like "N visualizations hidden because they're not supported on this device."

If you want a notification mechanism, I can add something — e.g. a dismissible info banner at the top of the view listing the hidden widget names. Or we could keep it silent as-is and rely on edit mode for discovery. What's your preference?

TimeToBuildBob:

ErikBjare:
@TimeToBuildBob add an option in settings to disable this hiding behavior, have it disabled by default so I can test it.

TimeToBuildBob:
Done — added `hideUnsupportedVisualizations` (default: `false`) in the latest commit.

**How to test:** go to Settings → Developer → "Hide unsupported visualizations" and toggle it on. The `v-if` in `SelectableVisualization.vue` now short-circuits when the setting is off, so the default experience is unchanged (shows the warning as before). Enabling the setting activates the hiding behavior from this PR.

PR #928, opened 2026-08-05
fix(queries): match Arc and Dia in the chrome appname patterns

### LINKED ISSUE #927: Browser view empty in Dia (Chromium fork, reports app name "Dia")
The Browser view (Top Domains / URLs / Titles) is permanently empty for me in [Dia](https://www.diabrowser.com/).  
The extension itself works fine, which is what made this annoying to track down.

Events land in `aw-watcher-web-chrome_<host>` and keep coming:

```json
{"timestamp": "2026-08-05T21:32:58Z", "duration": 20.1,
 "data": {"url": "http://localhost:5600/#/activity/...", "title": "ActivityWatch", "tabCount": 16}}
```

I'm pretty sure the problem is in `browserSummaryQuery`, which intersects those against window events whose `app` matches the chrome entry in `src/queries.ts`:

```ts
browser_appnames.chrome      = ['com.google.Chrome', 'com.google.ChromeDev', 'org.chromium.Chromium']
browser_appname_regex.chrome = '(?i)^(google[-_ ]?chrome|chrome|chromium)'
```

aw-watcher-window reports Dia as `app: "Dia"` (bundle id `company.thebrowser.dia`). But neither the exact list nor the regex matches that so `filter_period_intersect` doesn't return nothing and all three columns render "No data". 
Seems the same as #749 and #1094.

What I've confirmed fixes it:

```diff
-    chrome: ['com.google.Chrome', 'com.google.ChromeDev', 'org.chromium.Chromium'],
+    chrome: ['com.google.Chrome', 'com.google.ChromeDev', 'org.chromium.Chromium', 'company.thebrowser.dia'],
```
```diff
-    chrome: '(?i)^(google[-_ ]?chrome|chrome|chromium)',
+    chrome: '(?i)^(google[-_ ]?chrome|chrome|chromium|dia)',
```

I've been running that patched into the bundled webui inside the .app rather than a real build and it works.

Now the part I'm less sure about. My first instinct was that Dia should get its own key, the way `arc` does, since it is a separate browser. but I don't think that can work, and I think it's also why #1094 keeps collecting "still broken" replies despite `arc` having been added to `queries.ts`.

`browsersWithBuckets()` binds a browser key to a bucket by substring match on the bucket id:

```ts
const bucketId = _.find(browserbuckets, bucket_id => _.includes(bucket_id, browserName));
```

The extension can't tell it's running inside a fork, so Dia announces itself as chrome and its data goes to `aw-watcher-web-chrome`. 

But no bucket id ever contains "arc" or "dia", those keys get filtered out before their regexes are ever applied, and `arc: []` with `'(?i)^arc(\.exe)?$'` just sits there doing nothing. @wzxu noticed the same thing on #1094 ("Looks like Arc should already be defined in that file, but it's not working for me either"), doesn't look like it went anywhere.

Anyway, while the bucket name decides which key applies, forks gotta be listed under the browser whose bucket they actually write to. Adding `dia` to the chrome patterns works today. `arc` probably belongs there too. 

The other option could be have the extension declare which browser it is and binding on that, but that's a much bigger change and probably its own issue.

My local info:
macOS 15 (Darwin 25.5.0, Apple Silicon), ActivityWatch v0.14.0b3, Dia running the Chrome build of the extension.

TiberiusNemesis:
Fixes #927. Should also cover ActivityWatch/activitywatch#1094, which is the same failure mode with Arc.

As described in the issue: forks like Dia and Arc run the chrome build of the extension, so by default their web events land in `aw-watcher-web-chrome_<host>`, but aw-watcher-window reports them under their own app names ("Dia", "Arc"). Nothing in the chrome patterns matches those, so `filter_period_intersect` comes back empty and the Browser view shows no data.

One correction to what I wrote in the issue: I claimed a standalone key can never bind because no bucket id contains the fork's name. That's only true for the default setup. The extension settings actually have a browser dropdown (Arc is one of the options) that overrides the detected name, and picking it changes the bucket to `aw-watcher-web-arc_<host>`, which the existing `arc` key does bind to. So the standalone key stays for that case, and the chrome patterns pick up the default case.

So this PR:

- adds `arc` and `dia` alternatives to the chrome regex, each with its own terminal anchor so names like `archive` or `Dialog` don't match
- leaves the standalone `arc` entries alone, with a comment on the chrome entry explaining which case each one covers
- adds the fork names to the chrome pattern tests, plus negative cases for the anchoring

No `dia` key for now since the dropdown has no Dia option, so nothing ever produces a dia bucket. Easy to add alongside `arc` if the extension grows one.

I've been running this patched into the bundled webui inside the .app for a couple days and Top Domains/URLs/Titles populate fine in Dia now.

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -289,7 +289,12 @@ function browsersWithBuckets(browserbuckets: string[]): [string, string][] {
 // The full set of historical app names these patterns replace is documented in the unit tests.
 // See: test/unit/queries.test.node.ts, https://github.com/ActivityWatch/aw-webui/issues/749
 export const browser_appname_regex: Record<string, string> = {
-  chrome: '(?i)^(google[-_ ]?chrome|chrome|chromium)',
+  // Chromium forks (Arc, Dia) run the chrome build of the extension, which announces itself
+  // as chrome unless the user overrides the browser name in the extension settings. So by
+  // default their events land in the chrome bucket and their app names have to be matched
+  // here (#927, ActivityWatch/activitywatch#1094). The standalone arc key below only covers
+  // setups where Arc was picked explicitly in the settings, which changes the bucket name.
+  chrome: '(?i)^(google[-_ ]?chrome|chrome|chromium|arc(\\.exe)?$|dia(\\.exe)?$)',
   firefox: '(?i)(firefox|librewolf|waterfox|nightly)',
   opera: '(?i)(opera)',
   brave: '(?i)(brave)',
--- a/test/unit/queries.test.node.ts
+++ b/test/unit/queries.test.node.ts
@@ -36,6 +36,8 @@
  *             (Flatpak app IDs retained: 'com.microsoft.Edge', 'com.microsoft.EdgeDev')
  *
  *   Arc:      'arc.exe', 'Arc.exe', 'Arc'
+ *             (also matched by the chrome pattern, since Arc reports to the chrome bucket
+ *              unless the browser name is overridden in the extension settings)
  *
  *   Vivaldi:  'Vivaldi-stable', 'Vivaldi-snapshot', 'vivaldi.exe', 'Vivaldi.exe', 'Vivaldi'
  *             (Flatpak app ID retained: 'com.vivaldi.Vivaldi')
@@ -81,6 +83,12 @@ describe('browser_appname_regex', () => {
       'chromium.exe',
       'Google-chrome-beta',
       'Google-chrome-unstable',
+      // Chromium forks that report through the chrome extension bucket (#927)
+      'Arc',
+      'arc.exe',
+      'Arc.exe',
+      'Dia',
+      'Dia.exe',
     ];
     for (const name of knownNames) {
       expect(re.test(name)).toBe(true);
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Thanks for the diagnosis and the regex patch — that is the right shape.

I opened ActivityWatch/aw-webui#949 which takes the same chrome-regex approach (Dia/Arc `$`-anchored so `archive`/`Dialog` don't match, standalone `arc` key left for the settings-override case) and adds two gaps this PR doesn't cover yet:

1. `company.thebrowser.dia` on the exact chrome list (macOS bundle id; reverse-domain names are not matched by the process-name regex). That was in the original confirmed fix on #927.
2. An info hint in the Browser visualizations when a watcher bucket exists but the window intersection is empty, so the next fork is a visible miss instead of a silent "No data".

Happy to treat #949 as the follow-up on top of this PR instead, if you'd rather land this one first. No need for both to merge.

TimeToBuildBob:
Confirmed. ActivityWatch/aw-webui#949 was rebased onto the merged #928 and the overlap conflicts were resolved by keeping #928's landed regex/tests. The remaining follow-up is green and mergeable at `514a1fc`; its latest review is 5/5 with no blocking findings.

PR #933, opened 2026-08-13
fix(views): guard server info before dereferencing in templates

TimeToBuildBob:
## Problem

`useServerStore` starts with `info: null` and is only populated from `App.vue`'s
`mounted` hook — which runs *after* the first child render. Two templates
dereferenced `info` unguarded, so a cold load of either view throws:

```
TypeError: null is not an object (evaluating 't.info.version')
```

It renders as a red error banner over an otherwise perfectly working install.

Worse, when the server is unreachable the crash is **permanent**, not a
first-paint race: `getInfo()` catches the connection error and leaves `info`
null forever, so the banner never clears.

## Sites

| File | Expression | Status |
|---|---|---|
| `src/views/Home.vue:74` | `!info.version.includes('rust')` | fixed |
| `src/views/Buckets.vue:20` | `serverStore.info.hostname == device.hostname` | fixed |
| `src/components/Footer.vue:12,15` | `info.hostname` / `info.version` | already safe — ancestor `v-if="info"` |
| `src/stores/buckets.ts:44,187` | `serverStore.info && ...` | already safe |

## Regression test

`test/unit/serverInfoGuard.test.js` walks the pug template tree by indentation
and rejects any `info.*` dereference not covered by a guard on itself or an
ancestor. The indentation-awareness matters: a naive per-line check would
false-positive on Footer.vue, whose guard sits on the wrapping `span`. Covered
both ways — an ancestor guard passes, a sibling *after* a closed guard block
does not.

Verified against the pre-fix files from `master`: the detector flags both
offending lines and reports Footer clean, then reports all three clean after the
fix.

## Verification

- `npx jest --selectProjects jsdom` — 14 suites, 69 tests pass
- `make lint` clean on the changed files

## Origin

Found while triaging a participant's test report on the ActivityWatch Research
Edition build (`v0.14.0b3-research`). The install was working correctly and the
privacy filter was doing its job, but the banner made it look broken.

### DIFF
--- a/src/views/Buckets.vue
+++ b/src/views/Buckets.vue
@@ -17,7 +17,7 @@ div
         icon.mr-2.text-secondary(v-else name="desktop" scale="1.2")
         div
           span.font-weight-bold {{ device.hostname }}
-          b-badge.ml-2(v-if="serverStore.info.hostname == device.hostname" variant="info") {{ $t('buckets.thisDevice') }}
+          b-badge.ml-2(v-if="serverStore.info && serverStore.info.hostname == device.hostname" variant="info") {{ $t('buckets.thisDevice') }}
           div.small.text-muted(v-if="device.hostname !== device.device_id")
             | ID: {{ device.id }}
           div.small(v-if="deviceHasEvents(device)")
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:

TimeToBuildBob:

PR #942, opened 2026-08-22
fix: three small dashboard fixes (#232, #935, #299)

### LINKED ISSUE #232: [feature] Enter Key Handling in the category-edit modal
During my use of ActivityWatch I noticed, that the enter key in the category-edit modal does not work as expected (submitting the edit). I think it would be nice to have it do the same as pressing the "OK" button, because it really makes the process of editing and creating categories faster.
A quick google search showed, that it should only be necessary to add this argument
`@keydown.native.enter="handleOk"` to the b-modal in [CategoryEditTree](https://github.com/ActivityWatch/aw-webui/blob/master/src/components/CategoryEditTree.vue) (line 22). (From [here](https://github.com/bootstrap-vue/bootstrap-vue/issues/1509))
I didn't think a PR would be necessary for such a minor change, but I could just create one as well.

### LINKED ISSUE #299: Sticky navbar
It would be convenient for the navbar to always be present at the top of the window to prevent unneeded scrolling for tall pages.

The following additional CSS accomplishes this pretty well for desktop. Not sure how it works on mobile.

```css
nav.navbar {
    position: fixed;
    top: 0;
    width: 100%;
    z-index: 1;
}

/* the main container (may want to add extra specificity) */
.container {
    margin-top: 70px;
}
```

### LINKED ISSUE #935: Top Bucket Data visualization does not refresh when switching views
Switching between views does not refresh a Top Bucket Data visualization's content (selected bucket, field, and resulting events) if a visualization of the same type is already present at that position in another view.

0xbrayo:
Three unrelated, self-contained fixes for long-standing issues. Each is its own commit, so any one can be dropped without touching the others.

### `fix(categories)`: submit the category edit modal on Enter — closes #232

Pressing Enter in the category edit modal did nothing, so every category edit required reaching for the OK button. Enter now submits, skipping elements that handle it themselves (textarea, button, link, select), and respecting the same validity condition that disables OK.

### `fix(views)`: rebuild visualizations when switching dashboard views — closes #935

The visualization cards in `ActivityView` were keyed by their position in the list only, so switching to another view let Vue reuse the instance sitting at the same index in the previous view. Visualizations that keep local state kept the previous view's state — most visibly the Top Bucket Data picker, which holds its selected bucket, field and fetched events in `data` and populates them once in `mounted`, so it kept showing the old view's selection and events.

Keying on the view id as well forces a fresh instance per view.

### `fix(header)`: keep the navbar sticky on desktop too — closes #299

The navbar was pinned only on Android (`fixedTopMenu: this.$isAndroid`), so tall pages — the timeline, a long activity view — scrolled the navigation out of reach everywhere else. The `fixed-top-padding` rule that reserves space for the fixed navbar was already there; it was just gated on the platform.

This is the most opinionated of the three since it changes desktop layout for everyone, and it is the one I could not verify visually in CI. Happy to drop the commit if the preference is to make it a setting instead.

### Testing

- New unit tests for the Enter-key handling (valid/invalid pattern, and the tags that opt out).
- New regression test asserting the visualization key includes the view id; verified it fails on `master`.
- Full suite green: 227 tests, 33 suites. `lint` clean.

### DIFF
--- a/src/components/CategoryEditModal.vue
+++ b/src/components/CategoryEditModal.vue
@@ -1,6 +1,6 @@
 <template lang="pug">
 // The category edit modal
-b-modal(id="edit" ref="edit" title="Edit category" @show="resetModal" @hidden="hidden" @ok="handleOk" :ok-disabled="editing.rule.type === 'regex' && !validPattern")
+b-modal(id="edit" ref="edit" title="Edit category" @show="resetModal" @hidden="hidden" @ok="handleOk" @keydown.native.enter="handleEnter" :ok-disabled="editing.rule.type === 'regex' && !validPattern")
   div.my-1
     b-input-group.my-1(prepend="Name")
       b-form-input(v-model="editing.name")
@@ -132,6 +132,22 @@ export default {
         return this.validPattern;
       }
       return true;
+    },
+    handleEnter(event) {
+      // Enter submits the modal, same as pressing OK (#232).
+      // Skipped for elements where Enter already has a meaning of its own,
+      // so we don't swallow their default behavior.
+      const tag = event.target && event.target.tagName;
+      if (tag === 'TEXTAREA' || tag === 'BUTTON' || tag === 'A' || tag === 'SELECT') {
+        return;
+      }
+      // Mirrors the :ok-disabled condition on the modal
+      if (this.editing.rule.type === 'regex' && !this.validPattern) {
+        return;
+      }
+      event.preventDefault();
+      this.handleSubmit();
+      this.$emit('ok');
     },
     handleOk(event) {
       // Prevent modal from closing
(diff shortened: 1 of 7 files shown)


0xbrayo:
> 🤖 _Claude, on behalf of @0xbrayo_

Both review comments addressed in 8d37695:

- **Fixed navbar covers sticky settings** — the navbar height is now a `--aw-navbar-height` custom property shared by the navbar's own spacer padding and the settings sidebar, which sticks at `calc(var(--aw-navbar-height) + 1rem)` so it clears the navbar.
- **Test checks key source text** — replaced with a mount test that switches `view_id` and asserts a fresh `aw-selectable-vis` instance is created (plus the negative case: no rebuild when the view is unchanged). Verified it fails with the old index-only key.

@greptile review

PR #946, opened 2026-08-24
fix(layout): preserve bottom margin on small screens

### LINKED ISSUE #236: Empty changelog + versioning UX issues + minor nits after v0.14.0b2 release to stable channel
The just released v0.14.0b2 on Play Store includes a short summary of changes which links to a "full changelog" which was (1) a draft unpublished release, invisible to others likely and (2) empty description. I was expecting a similarly generated release description as used for the main ActivityWatch and gptme repos (should probably share/copy their script with reference or wget it from those repos + run in CI job like they do).

Also, the bottom "Version: v0.14.0 (rust)" in the webui is misleading (should be `v0.14.0b2` not what it happened to say in the Cargo.toml for aw-server-rust, it should be overridden with the *release version* not the component version - which doesn't even include a submodule commit or anything so it's not that precise anyway)

The webui is missing some margin/padding right above the "Made with <3 by the ActivityWatch developers", at least on small screens. Prob due to the `my-sm-3` class which should probably also have `mb-3` to fix so the bottom margin remains (the top should indeed go).

@TimeToBuildBob Go

TimeToBuildBob:
Fixes the missing bottom margin above the footer content on small screens.

The `my-sm-3` class only applies vertical margin on `sm`+ breakpoints, leaving no bottom margin on extra-small screens. Adding `mb-3` preserves the bottom spacing on all screen sizes while the top margin removal on small screens (the intent of `my-sm-3` alone) is maintained.

**Before**: no bottom padding above footer "Made with ❤" on small screens
**After**: consistent `mb-3` bottom spacing on all screen sizes

Closes ActivityWatch/aw-android#236 (footer margin nit)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -3,7 +3,7 @@ div#wrapper(v-if="loaded")
   aw-header
 
   div(:class="{'container': !fullContainer, 'container-fluid': fullContainer}").px-0.px-md-2
-    div.aw-container.my-sm-3.p-3
+    div.aw-container.my-sm-3.mb-3.p-3
       error-boundary
         user-satisfaction-poll
         new-release-notification(v-if="isNewReleaseCheckEnabled")


TimeToBuildBob:

Issue #925, opened 2026-08-03
Improve the single-pass activity-analysis context


TimeToBuildBob:
## Context

Follow-up from ActivityWatch/aw-webui#922. The dev-mode `/analysis/activity` prototype currently sends only total tracked duration plus the top 20 applications to a single-pass LLM. Because that model has no ActivityWatch tools, the compact context needs to carry enough structure for a useful answer without sending raw personal event data.

## Proposed context shape

Build on the existing canonical-query / aggregation seam and include:

- active time after AFK filtering, plus tracked-vs-active coverage
- top applications with duration, share, and distinct-title count
- top browser domains with duration and share (when browser URL data exists)
- category rollups using the user's ActivityWatch categorization rules
- daily distribution (per-day active time) and focus fragmentation signals such as app switches / longest sustained block
- explicit date range, host(s), timezone, and truncation/coverage notes

Keep each section bounded and inspectable in the UI before generation. Derive statistics locally; do not send raw titles or URLs by default.

## Acceptance criteria

- [ ] Define a typed provider-independent summary input shape
- [ ] Fetch through ActivityWatch's query infrastructure rather than downloading an uncapped raw bucket where practical
- [ ] Apply AFK filtering and user category rules
- [ ] Add top domains and title-count metadata without exposing full URLs/titles
- [ ] Add bounded temporal/focus statistics with clear definitions
- [ ] Show the exact compact context sent to the provider
- [ ] Add unit tests for aggregation, missing buckets/rules, percentage denominators, and truncation
- [ ] Document which fields may contain sensitive information

This should be a follow-up, not scope-creep in ActivityWatch/aw-webui#922.

ErikBjare:
I also want a way for the user to query with a privacy-aware/scrubbed subset (exclude uncategorized, exclude private/sensitive categories).

TimeToBuildBob:
Agreed. I'll make the compact-context preview support an explicit privacy filter before generation: exclude uncategorized activity and exclude categories marked private/sensitive, with the exclusions and resulting coverage visible so the user can verify what leaves the device. Raw titles and URLs will remain excluded by default.

I'll treat privacy classification as user-controlled category metadata rather than hard-coding category names; the implementation slice needs to define that metadata/UX alongside the filtered query path.

TimeToBuildBob:
[...]

PR #948, opened 2026-08-26
feat(ai-summary): typed, privacy-filtered activity context (#925)

TimeToBuildBob:
Follow-up to #922, implementing ActivityWatch/aw-webui#925.

The dev-mode `/analysis/activity` page sent a single-pass LLM only total tracked
duration plus a flat top-20 app list, built from an uncapped raw bucket download.
This replaces that with a typed, bounded context derived through the query layer.

## What's in the context

`src/util/activityContext.ts` defines a provider-independent `ActivityContext`
and pure builders over it (no store/query imports, so it's node-testable):

- **coverage** — tracked time vs active time after AFK filtering, plus the time
  actually exported after privacy filtering
- **categories** — rollups using the user's own category rules
- **apps** — duration, share, and a distinct-title *count* (titles stay local)
- **domains** — browser domains, host only; never paths or query strings
- **daily** — active time per day in the user's timezone
- **focus** — app switches, block count, longest and median uninterrupted block

`analysisContextQuery()` in `queries.ts` does AFK filtering and categorization
server-side and returns the timeline, browser domains, and the unfiltered
tracked duration — instead of `getEvents(..., {limit: -1})`.

## Privacy

@ErikBjare asked in #925 for "a way to query with a privacy-aware/scrubbed
subset (exclude uncategorized, exclude private/sensitive categories)". Both are
implemented as opt-in checkboxes:

- **Exclude uncategorized activity** — drops anything that matched no rule
- **Exclude categories marked private** — driven by *user-controlled category
  metadata* (`data.private === true`), not a hard-coded name list, so it
  survives renames and covers categories we've never heard of. Sub-categories
  of a private category are excluded too (`['Private','Messaging']` under
  `['Private']`, while `['Workout']` is *not* under `['Work']`).

The context reports exactly how much was withheld (`privacy.excludedSeconds`,
`privacy.coverage`) and renders it in the text, so the user can verify what
left the device before generating.

Browser domains are dropped entirely while either filter is active: browser
events carry no `$category`, so they can't be filtered by one, and leaking them
would defeat the filter the user just enabled.

## Fields that may contain sensitive information

Documented in the module header:

| Exported | Sensitivity |
|---|---|
| `apps[].app` | application names (e.g. a therapy or banking app) |
| `domains[].domain` | browser domains, host only |
| `categories[].category` | the user's own category names |

Never exported: window titles, full URLs, event timestamps. Titles are reduced
to a per-app distinct count.

## Truncation

Every bounded list reports `{shown, total, otherSeconds}` and renders a
`(+N more, Xh Ym)` line, so a truncated section is visible to the model rather
than silently partial.

## Acceptance criteria

- [x] Typed provider-independent summary input shape
- [x] Fetch through query infrastructure rather than an uncapped raw bucket
- [x] AFK filtering and user category rules
- [x] Top domains and title-count metadata without full URLs/titles
- [x] Bounded temporal/focus statistics with clear definitions
- [x] Show the exact compact context sent to the provider (the "Show context sent" card)
- [x] Unit tests for aggregation, missing buckets/rules, percentage denominators, truncation
- [x] Document which fields may contain sensitive information
- [x] Privacy filter (in-thread ask)

## Testing

33 new unit tests in `test/unit/activityContext.test.node.ts`; `tsc --noEmit`
clean; eslint clean. The superseded `aggregateEvents`/`buildSummaryText`
helpers and their tests are removed — nothing else referenced them.


TimeToBuildBob:
Addressed the five P1 issues from the Greptile review (commit 415e985):

**1. Empty `filter_categories: []` zeroed all events** — `params.filter_categories` is truthy for `[]` in JS, so `filter_keyvals(events, "$category", [])` was keeping only events whose `$category` is in the empty allow-list (zero events). Fixed in `queries.ts`: now skips the `filter_keyvals` statement when the array is empty.

**2. Missing AFK bucket caused query failure** — When no AFK bucket exists, `bid_afk: ''` still generated `query_bucket("")` which the server rejects. Fixed in `canonicalEvents`: guarded on `params.bid_afk` being non-empty; emits `not_afk = [];` as a safe fallback when there's no AFK bucket.

**3. Browser buckets crossed hosts** — `buckets.filter(b => b.type === 'web.tab.current')` collected browser buckets from *all* hosts, so another device's domains could appear in a single-host summary. Fixed in `AISummaryView.vue`: added `&& b.hostname === this.selectedHost` to the filter.

**4. Excluded private category names leaked to the LLM** — `formatActivityContext` included `excludedCategories.map(c => c.join(' > ')).join(', ')` in the provider-bound text, exposing names like "Health" or "Finance" despite those categories being excluded. Fixed to emit only the count (`N excluded categories`).

**5. Events spanning midnight misattributed** — Duration was fully assigned to the start day. Added `accumulateByDay()` which uses a binary-search midnight finder to split events at local day boundaries; the daily loop now calls this instead of `dayKey()` directly.

Tests updated accordingly; all 35 unit tests pass.

src/util/activityContext.ts
```diff
@@ -0,0 +1,498 @@
...
+  );
+  lines.push(`Summarized below: ${formatDuration(ctx.coverage.exportedSeconds)}`);
+
+  if (ctx.privacy.excludedSeconds > 0 || ctx.privacy.excludedCategories.length > 0) {
+    lines.push('');
+    lines.push(
+      `Privacy filter: withheld ${formatDuration(ctx.privacy.excludedSeconds)} ` +
+        `(${formatShare(1 - ctx.privacy.coverage)} of active time) across ` +
+        ctx.privacy.excludedCategories.map(c => c.join(' > ')).join(', ')
+    );
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 415e985: `formatActivityContext` now emits only the count of excluded categories (e.g. `1 private category excluded`) rather than the category names themselves. The path string is never included in the provider-bound text.

TimeToBuildBob:
@greptileai review

src/views/AISummaryView.vue
```diff
@@ -210,19 +227,48 @@ export default {
...
-      return getClient().getEvents(windowBucket.id, { start, end, limit: -1 });
+      // Query-side AFK filtering and categorization: only derived statistics
+      // (never raw titles or URLs) are exported from the result below.
+      const query = analysisContextQuery({
+        bid_window: windowBucket.id,
+        bid_afk: afkBucket ? afkBucket.id : '',
+        bid_browsers: browserBuckets,
+        filter_afk: Boolean(afkBucket),
+        categories: this.categoryStore.classes_for_query,
+        filter_categories: [],
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 415e985: when `filter_categories` is an empty array, the query now skips `filter_keyvals` entirely (no-op), so all events pass through. Previously `filter_keyvals([...])` with an empty allow-list removed everything.

src/views/AISummaryView.vue
```diff
@@ -210,19 +227,48 @@ export default {
...
       const start = new Date(end.getTime() - this.periodDays * 24 * 60 * 60 * 1000);
 
-      return getClient().getEvents(windowBucket.id, { start, end, limit: -1 });
+      // Query-side AFK filtering and categorization: only derived statistics
+      // (never raw titles or URLs) are exported from the result below.
+      const query = analysisContextQuery({
+        bid_window: windowBucket.id,
+        bid_afk: afkBucket ? afkBucket.id : '',
+        bid_browsers: browserBuckets,
+        filter_afk: Boolean(afkBucket),
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 415e985: when `bid_afk` is empty (no AFK bucket for this host), the query now emits `not_afk = [];` as a safe empty result rather than `query_bucket("")`, which was rejected by the server.

src/views/AISummaryView.vue
```diff
@@ -210,19 +227,48 @@ export default {
...
       const windowBucket = buckets.find(
         b => b.type === 'currentwindow' && b.hostname === this.selectedHost
       );
       if (!windowBucket) {
         throw new Error(`No window-watcher bucket found for host: ${this.selectedHost}`);
       }
+      const afkBucket = buckets.find(
+        b => b.type === 'afkstatus' && b.hostname === this.selectedHost
+      );
+      const browserBuckets = buckets.filter(b => b.type === 'web.tab.current').map(b => b.id);
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 415e985: the browser bucket filter is now `.filter(b => b.hostname === selectedHost)`, so only buckets belonging to the currently selected host are included. Cross-host domains are no longer sent.

src/util/activityContext.ts
```diff
@@ -0,0 +1,498 @@
...
+  const domainTotal = Object.values(domainDurations).reduce((sum, d) => sum + d, 0);
+  const allDomains: DomainStat[] = Object.entries(domainDurations)
+    .map(([domain, duration]) => ({ domain, duration, share: share(duration, domainTotal) }))
+    .sort((a, b) => b.duration - a.duration);
+
+  // Daily distribution.
+  const dayDurations: Record<string, number> = {};
+  for (const event of kept) {
+    const key = dayKey(event.timestamp, input.timezone);
+    dayDurations[key] = (dayDurations[key] || 0) + (event.duration || 0);
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 415e985: added `accumulateByDay()` which splits event durations at midnight boundaries (via binary-search on the timezone offset transition) before accumulating, so post-midnight activity is correctly attributed to the next day. Added a regression test covering midnight-spanning events.

TimeToBuildBob:
Fixed the remaining CI failures (commit a00f383):

**1. Lint: non-null assertion warnings** — `activityContext.test.node.ts` used `aug1!.duration` after `expect(aug1).toBeDefined()`, which eslint correctly flags. Restructured the midnight-split test to use `Object.fromEntries` so TypeScript narrows the type without `!`.

**2. Test snapshot mismatch** — `queries.test.node.js.snap` still contained `events = filter_keyvals(events, "$category", true);` which was legitimately removed by the P1 fix (skip that statement when `filter_categories` is empty). Updated the snapshot to match the corrected query.

TimeToBuildBob:
## Greptile convergence adjudication — merge recommendation

**Context**: This PR hit the Greptile attempt cap (2 review rounds). Adjudicating the remaining P1 finding inline rather than triggering another review.

---

### Fixed this session

| Finding | File | Fix |
|---|---|---|
| P1: Browser domains omitted | `src/views/AISummaryView.vue:245` | Replaced inline `web.tab.current` filter (strict hostname equality) with `this.bucketsStore.bucketsBrowser(this.selectedHost)` — the existing store getter that already implements `fallback_unknown_host: true`, so browser buckets stored under the `'unknown'` hostname convention are included. Commit: d8b3873 |

**Root cause**: The inline filter `b.hostname === this.selectedHost` excluded browser buckets whose hostname is `'unknown'` — the established AW convention for watcher-side unresolved hosts. The rest of the codebase already handled this via `bucketsByType(host, type, fallback_unknown_host: true)`, but `buildContextText()` bypassed the store getter and reimplemented the filter without the fallback.

**Tests**: All 39 existing unit tests pass against the fix (`aiSummary` and `activityContext` suites). Existing `bucketsBrowser` coverage in the store handles the unknown-hostname path.

---

### Remaining findings

All prior P1 findings from round 1 (midnight interval allocation, private-name exposure, empty-filter zeroing events, missing AFK bucket, cross-host browser buckets) were resolved in commit 415e985 and confirmed resolved by the round-2 review.

No P2 or P3 findings remain unresolved.

---

### CI

CI is running against the new head (d8b3873). Prior CI passed on a00f383 (three commits before the fix). The change is a one-line substitution with no query-side impact.

---

### Domain risk

The `buildContextText()` path is only triggered when a user clicks "Generate summary" and has the feature configured. The change is strictly additive: users with a correctly-hostnamed browser bucket see no difference; users with an `unknown`-hostname browser bucket now get their domain activity included in the summary instead of silently omitted. No data is removed or reordered.

The `bucketsBrowser()` getter is already exercised on every AISummaryView load (for the host-availability check), so this is not a new code path.

---

### Convergence

`round_convergence.stable_rounds = 0`, `new_blocking_count = 1` (the browser-domains finding, now fixed). The cap was reached after 2 rounds; this session adjudicated and fixed the remaining blocking item rather than requesting a third Greptile pass.

---

**Status**: Ready for maintainer review and merge. No open blocking findings. CI pending on d8b3873.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the remaining Greptile P1 (commit d8b3873):

**"Browser domains omitted on `unknown`-hostname buckets"** — the strict `b.hostname === this.selectedHost` filter silently dropped browser buckets stored under the `unknown` hostname (the established AW convention when the watcher can't resolve the host). Fixed by switching to `this.bucketsStore.bucketsBrowser(this.selectedHost)`, which already implements the `fallback_unknown_host` logic used by all other bucket getters in the store.

Greptile re-review triggered on the current HEAD. Old review thread resolved. CI running.

TimeToBuildBob:

src/views/AISummaryView.vue
```diff
@@ -210,19 +227,53 @@ export default {
...
         throw new Error(`No window-watcher bucket found for host: ${this.selectedHost}`);
       }
+      const afkBucket = buckets.find(
+        b => b.type === 'afkstatus' && b.hostname === this.selectedHost
+      );
+      // Only include browser buckets for the selected host; mixing buckets from
+      // other hosts would send another device's domains to the LLM under the
+      // selected host's summary.
+      const browserBuckets = buckets
+        .filter(b => b.type === 'web.tab.current' && b.hostname === this.selectedHost)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in d8b3873: switched from the inline `b.hostname === this.selectedHost` filter to `this.bucketsStore.bucketsBrowser(this.selectedHost)`. The store getter already implements `fallback_unknown_host: true`, so browser buckets stored under the `unknown` hostname convention are now included. Greptile round 3 came back clean.

TimeToBuildBob:
Rebased onto current `master` and force-pushed the conflict-free history in `ad72b2d`.

Conflict resolution preserved both sides of the overlapping query change: this PR's missing-AFK/category-filter handling and the now-landed per-app desktop merge behavior. Verified locally with `tsc --noEmit`, ESLint on the changed files, and the 59 focused `activityContext`, `aiSummary`, and `queries` tests.

TimeToBuildBob:
## 🤖 AI code review

This PR replaces the AI summary view's raw bucket download with a typed, privacy-filtered activity context. It adds analysisContextQuery() in queries.ts, a new activityContext.ts module with buildActivityContext/formatActivityContext, and rewires AISummaryView.vue to use them. It also removes the old aggregateEvents/buildSummaryText helpers and their tests, and adds a new test file for activityContext.

### Safe to merge — no P0/P1 findings

Confidence 5/5

✅ **No findings.** The diff looks correct to me on this pass.

<details><summary><b>Files changed (7)</b> — the diff as I read it</summary>

- `src/queries.ts` — Adds analysisContextQuery() and changes canonicalEvents to emit an empty not_afk list when bid_afk is empty and to skip filter_keyvals when filter_categories is empty.
- `src/util/activityContext.ts` — New module defining ActivityContext types, buildActivityContext(), computeFocusStats(), formatActivityContext(), and privacy helpers.
- `src/util/aiSummary.ts` — Removes aggregateEvents, formatDurationHuman, buildSummaryText and their interfaces; keeps LLM config and callLLM.
- `src/views/AISummaryView.vue` — Adds privacy checkboxes, uses analysisContextQuery and buildActivityContext, filters Android buckets from host options, and loads category store.
- `test/unit/__snapshots__/queries.test.node.js.snap` — Updates fullDesktopQuery snapshot to remove the empty filter_keyvals line.
- `test/unit/activityContext.test.node.ts` — New tests for buildActivityContext, computeFocusStats, formatActivityContext, and privacy helpers.
- `test/unit/aiSummary.test.node.ts` — Removes tests for the deleted aggregateEvents/buildSummaryText functions.

</details>

<details><summary>Previous review passes</summary>

| commit | score | findings | engine | when |
| --- | --- | --- | --- | --- |
| `ad72b2d8202e` | 1/5 | 5 | agent | 2026-09-16 09:00 UTC |

</details>

<sub>Reviewed `9e8416ee000c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 17s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

src/views/AISummaryView.vue
```diff
@@ -210,19 +227,54 @@ export default {
...
+      const afkBucket = buckets.find(
+        b => b.type === 'afkstatus' && b.hostname === this.selectedHost
+      );
+      // Only include browser buckets for the selected host; mixing buckets from
+      // other hosts would send another device's domains to the LLM under the
+      // selected host's summary. Use bucketsBrowser() rather than an inline
+      // filter so the established 'unknown'-hostname fallback is respected: when
+      // a browser bucket's hostname is 'unknown' (common in older AW setups),
+      // strict equality against this.selectedHost would silently drop it.
+      const browserBuckets = this.bucketsStore.bucketsBrowser(this.selectedHost);
```
TimeToBuildBob, line 257: 🛑 **P0** — `bucketsBrowser()` falls back to every browser bucket whose hostname is `unknown` when the selected host has no exact match. In an older multi-device installation, those buckets cannot be attributed to the selected host, so domains from another device can be included in the context and sent to an external LLM despite the UI presenting this as a single-host summary. Because domains are sensitive data, ambiguous buckets should be omitted or require explicit user selection/consent.

src/views/AISummaryView.vue
```diff
@@ -210,19 +227,54 @@ export default {
         this.loading = false;
       }
     },
-    async fetchEvents(): Promise<any[]> {
+    async buildContextText(): Promise<string> {
       const buckets = this.bucketsStore.buckets || [];
       const windowBucket = buckets.find(
```
TimeToBuildBob: ❌ **P1** — This selects any `currentwindow` bucket rather than using the store's `bucketsWindow()` filtering. `aw-watcher-android` also uses the `currentwindow` type and sorts before `aw-watcher-window`, so a host containing both can silently summarize the Android bucket and then intersect it with the desktop AFK bucket, often yielding incorrect or empty results. Android-only hosts are also offered even though this builds a desktop query. Filter out Android buckets or select through `bucketsWindow(selectedHost)`.

src/views/AISummaryView.vue
```diff
@@ -210,19 +227,54 @@ export default {
...
+      // (never raw titles or URLs) are exported from the result below.
+      const query = analysisContextQuery({
+        bid_window: windowBucket.id,
+        bid_afk: afkBucket ? afkBucket.id : '',
+        bid_browsers: browserBuckets,
+        filter_afk: Boolean(afkBucket),
+        categories: this.categoryStore.classes_for_query,
+        filter_categories: [],
+      });
+      const [result] = await getClient().query([{ start, end }], query);
```
TimeToBuildBob, line 277: ⚠️ **P2** — The 7/30/90-day range is submitted as one query and `analysisContextQuery` returns every detailed window event without a limit. This remains an uncapped raw timeline download and, especially for the 90-day option, can produce a very large response or exceed the client's default 30-second timeout. Other desktop-query paths explicitly split multi-day ranges to avoid this failure mode; this path should similarly chunk requests or aggregate the required statistics server-side.

src/util/activityContext.ts
```diff
@@ -0,0 +1,551 @@
...
+ * than watcher heartbeats.
+ */
+export function computeFocusStats(events: ContextEvent[]): FocusStats {
+  const blocks: number[] = [];
+  let currentApp: string | null = null;
+  let currentDuration = 0;
+
+  for (const event of events) {
+    const app = event.data?.app || 'unknown';
+    if (app === currentApp) {
```
TimeToBuildBob: ⚠️ **P2** — Focus blocks are merged solely by adjacent app name. After AFK or privacy filtering, two events for the same app can be adjacent in this array while separated by minutes or hours, yet their durations are combined and reported as one “uninterrupted” block. Block construction needs to compare each event's timestamp with the previous event's end and start a new block when there is a gap.

src/queries.ts
```diff
@@ -490,6 +494,34 @@ export function activityQueryAndroid(androidbucket: string): string[] {
...
+    ${canonicalEvents({
+      ...params,
+      bid_window: escape_doublequote(params.bid_window),
+      bid_afk: escape_doublequote(params.bid_afk),
+      bid_browsers: _.map(params.bid_browsers, escape_doublequote),
+    })}
+    events = sort_by_timestamp(events);
+    browser_events = split_url_events(browser_events);
+    browser_domains = sort_by_duration(merge_events_by_keys(browser_events, ["$domain"]));
+    browser_domains = limit_events(browser_domains, ${default_limit});
```
TimeToBuildBob: ⚠️ **P2** — Domains are capped at 100 before `buildActivityContext` computes its truncation metadata. Consequently, for more than 100 domains, `truncation.domains.total` can never report the real total and `otherSeconds` omits all time outside the server-side top 100, contradicting the context's claim that bounded lists describe what was dropped. Return an aggregate remainder/count from the query or defer the limit until the client has computed the metadata.

src/views/AISummaryView.vue
TimeToBuildBob: Fixed in 9e8416e (plus d6493a8): the fallback is now kept (dropping it loses all browser data on older single-device setups where the browser watcher stores hostname `unknown`) but is explicitly disclosed: `browserFallbackUsed` detects when fallback buckets are included (not just when no exact-host buckets exist), and the context text is prefixed with a NOTE that those domains may belong to a different device. The user reviews the exact context in the preview before it is sent to the LLM, so the disclosure is visible in the bytes that actually leave the machine.

src/views/AISummaryView.vue
TimeToBuildBob: Fixed in 5665dc1: window bucket selection now goes through the store's `bucketsWindow(selectedHost)` filter (excludes `aw-watcher-android` buckets), and the host picker only offers hosts with a non-Android `currentwindow` bucket.

src/util/activityContext.ts
TimeToBuildBob: Fixed in 5665dc1: `computeFocusStats` now tracks each event's end timestamp and starts a new block when the next same-app event starts more than 1s after the previous end (watcher-heartbeat jitter tolerance). Added a regression test for the gap case.

src/queries.ts
TimeToBuildBob: Fixed in 5665dc1: `analysisContextQuery` no longer caps `browser_domains` server-side. Domain events are one short string per domain, so the response stays small, and `buildActivityContext` now computes `truncation.domains` over the full list, so `total` and `otherSeconds` are accurate.

src/views/AISummaryView.vue
TimeToBuildBob: Accepted trade-off for this PR. The events returned here are the AFK-filtered, categorized timeline — titles and URLs never leave the device (see the comment block on `analysisContextQuery`), and only derived, bounded statistics are exported from them. Chunking the range like `desktopQuerySplit` does is a real improvement, but it changes the query/merge layer this PR deliberately kept minimal; I've noted it as follow-up work rather than expanding this PR's scope.

PR #965, opened 2026-09-03
fix(categories): enable Save after switching category set (closes #955)

### LINKED ISSUE #955: Category set dropdown doesn't apply selection or prompt save
## Bug

Switching the category set via the `<select>` dropdown has no effect.

## Steps to reproduce

1. Open the "Category set" dropdown
2. Select a different category set from the list

## Expected behavior

The app should either apply the selected set of categories immediately or prompt the user to save it.

## Actual behavior

Nothing happens. The set of categories changes visually, but it isn't actually applied, and the save prompt doesn't appear.

## Workaround

After switching the set, manually add and delete a category. This triggers the unsaved changes alert and enables the "Save" button, allowing the user to save and apply the new category set.

TimeToBuildBob:
## Problem

When switching between category sets in Categorization Settings, the **Save button stays disabled** and the selection change is never persisted. A page refresh silently reverts to the previous set.

Root cause: `switchToSet()` always set `classes_unsaved_changes = false` after switching, even though the new `active_set_ids` value hadn't been written to storage yet. The Save button is gated on `classes_unsaved_changes`, so it stayed disabled.

## Fix

In `src/stores/categories.ts`, `switchToSet()` now:
- Detects whether the active set actually changed (avoids spurious dirty state when re-selecting the same set or initializing)
- Sets `classes_unsaved_changes = true` when the set changed, activating the Save button

Two regression tests added to `test/unit/store/categories.test.node.ts`:
- Switching to a different set marks dirty ✓  
- Re-selecting the already-active set does not mark dirty ✓

## Testing

```
npx jest test/unit/store/categories.test.node.ts  # 12/12 pass
npx jest test/unit/presetCategories.test.node.ts  # 34/34 pass
npx tsc --noEmit                                   # clean
```

Closes #955

### DIFF
--- a/src/stores/categories.ts
+++ b/src/stores/categories.ts
@@ -263,10 +263,16 @@ export const useCategoryStore = defineStore('categories', {
         console.warn('Category set not found:', id);
         return;
       }
+      // Track whether the active set actually changed so we can mark dirty only
+      // when there is something to save (avoids spurious unsaved-changes prompts
+      // when initialising or re-selecting the already-active set).
+      const changed = this.active_set_ids.length !== 1 || this.active_set_ids[0] !== id;
       syncToPrimarySet(this);
       this.active_set_ids = [id];
       this.classes = computeEffectiveClasses(this.category_sets, this.active_set_ids);
-      this.classes_unsaved_changes = false;
+      // When the active set changed, mark unsaved so the Save button activates
+      // and the user can persist the new active_set_ids to storage.
+      this.classes_unsaved_changes = changed;
     },
 
     /**
--- a/test/unit/store/categories.test.node.ts
+++ b/test/unit/store/categories.test.node.ts
@@ -196,4 +196,37 @@ describe('categories store', () => {
     categoryStore.save();
     expect(categoryStore.classes_unsaved_changes).toBeFalsy();
   });
+
+  test('switchToSet marks dirty when switching to a different set', () => {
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
The failed `aw-server-rust master` job was a CI cleanup failure, not a test failure: all 244 unit tests and 7 e2e tests passed, then both log steps errored because no files matched `~/.cache/activitywatch/log/*/*.log`.

I pushed `b8bb1d3`, making those cleanup steps no-ops when logs are absent. The replacement run is now fully green (including all three test-matrix jobs, builds, lint, and CodeQL).

TimeToBuildBob:

TimeToBuildBob:
Rebased onto current `master` and force-pushed `6bb8ff0`. The only conflict was the PR's temporary CI log-cleanup commit versus the now-merged upstream fix in ActivityWatch/aw-webui#962; I dropped the redundant PR commit and kept upstream's broader `activitywatch*` cache-path handling.

The category fix itself rebased cleanly. Verified locally: category-store tests pass (12/12), TypeScript checks clean, and `git diff --check` is clean. Fresh CI is starting on the rebased head.

ErikBjare:
@TimeToBuildBob Resolve conflict

TimeToBuildBob:
Conflict resolved — rebased onto master (which added the `import syncs classes` test in #956). Kept both test additions. CI running on new head `e95fc56`.

PR #969, opened 2026-09-04
fix(activity): use actual bucket IDs in multidevice query (fixes BucketNotFound with aw-sync data)

Q-Ze:
# Fix BucketNotFound in multidevice query when viewing aw-sync data

## Problem

With `aw-sync` set up between machines, enabling the "Use multidevice query" developer setting and opening the Activity view fails on every machine except the one being viewed:

```
BucketNotFound("Failed to find bucket 'aw-watcher-window_<hostname>'")
```

## Root cause

Buckets pulled in by `aw-sync` keep their original `hostname` **field**, but their bucket **ID** gets an `-synced-from-<host>` suffix (e.g. `aw-watcher-window_myhost-synced-from-myhost`).

- `bucketsByHostname` groups by the hostname field, so synced hosts correctly appear in the device list;
- but `multideviceQuery` reconstructs bucket IDs from the hostname (`get_params`: `bid_window: 'aw-watcher-window_' + host`) and `query_multidevice_full` passes `host_params: {}`, so the existing per-host override mechanism in `get_params` is never used;
- `query_bucket("aw-watcher-window_myhost")` then fails with `BucketNotFound`, since only the suffixed ID exists in the local datastore.

## Fix

1. Fill `host_params` with each host's actual bucket IDs (first window/afk bucket from the buckets store), so synced buckets are queried by their real IDs.
2. Only include hosts that have both a window and an afk bucket (the pair `canonicalEvents` requires), matching the single-device path — previously only the window bucket was required, which could produce the same error for the afk bucket.

## Verification

With two machines synced via aw-sync (one viewing the other's data):

- **Before:** the query the UI generates returns HTTP 400 `BucketNotFound("Failed to find bucket 'aw-watcher-window_QBZdeMac-mini.local'")`.
- **After:** the same multidevice query (two hosts, `union_no_overlap`) executes successfully, e.g. `{"duration": 22392.738}` for the day across both hosts.

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -54,7 +54,7 @@ interface Rule {
 
 type Category = [string[], Rule];
 
-interface BaseQueryParams {
+export interface BaseQueryParams {
   include_audible?: boolean;
   categories: Category[];
   filter_categories: string[][];
@@ -63,14 +63,14 @@ interface BaseQueryParams {
   return_variable_suffix?: string;
 }
 
-interface DesktopQueryParams extends BaseQueryParams {
+export interface DesktopQueryParams extends BaseQueryParams {
   bid_window: string;
   bid_afk: string;
   filter_afk: boolean;
   always_active_pattern?: string;
 }
 
-interface AndroidQueryParams extends BaseQueryParams {
+export interface AndroidQueryParams extends BaseQueryParams {
   bid_android: string;
   /** True when the bucket is an aw-import-screentime (iOS) bucket.
    *  ScreenTime events carry a "title" key; aw-watcher-android events do not.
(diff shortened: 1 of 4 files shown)


Q-Ze:
Thanks for the review! Both non-blocking findings are addressed in d5ab9e7:

- **Test coverage**: the host-selection logic now lives in a pure helper (`buildMultideviceHostParams`, `src/util/multidevice.ts`) with unit tests, plus a regression test asserting the generated query references the actual (suffixed) bucket IDs and not the reconstructed `aw-watcher-window_<host>` ones.
- **Type safety**: the `as any` cast is gone — `host_params` values are now typed as `Partial<DesktopQueryParams> | Partial<AndroidQueryParams>`, which matches the override semantics in `get_params` (only keys present and non-empty are applied).

ErikBjare:
Reviewed in the context of the sync work tracked in ActivityWatch/activitywatch#1445 — this is the right fix for the current model, and it is directionally the change proposed in ActivityWatch/activitywatch#302 (stop reconstructing bucket IDs from hostnames in the frontend; query the IDs the store actually has).

Verified: `buildMultideviceHostParams` uses the store's real per-host bucket lists, no prefix matching is added, the window+afk pairing matches what `canonicalEvents` requires, and the tests cover the `-synced-from-` case that produced the `BucketNotFound`. The `[0]` pick when a host has two `currentwindow` buckets is pre-existing behaviour from the single-device path, not something this PR introduces.

Thanks for the clear root-cause write-up — it is the same hostname-as-identity problem the sync issues keep hitting from the other side. Recommending merge.

PR #970, opened 2026-09-05
fix(timeline): handle synced buckets where origin hostname contains underscores

### LINKED ISSUE #967: Timeline: synced bucket labels truncate the origin hostname at the first underscore
### Summary

Timeline row labels cut the origin hostname of `-synced-from-` buckets at the first
underscore, so `aw-watcher-android-synced-from-my_phone` renders as
`android-synced-from-my`.

### Steps to reproduce

1. Import a bucket (Buckets → Import buckets), timestamp set to a day you're viewing:

```json
{ "buckets": { "b": {
  "id": "aw-watcher-android-synced-from-my_phone",
  "type": "currentwindow", "client": "aw-android", "hostname": "my_phone",
  "events": [ { "timestamp": "2026-09-03T10:00:00Z", "duration": 600,
                "data": { "app": "Firefox" } } ] } } }
```

2. Open Timeline for that day.

**Actual:** row label is `android-synced-from-my`.
**Expected:** `android (synced from my_phone)`, as #757 produces for desktop buckets.

### Cause

`src/util/timelineLabels.ts`. `shortenBucketLabel` drops everything from the first
underscore, which on desktop separates the host
(`aw-watcher-window_erb-m2.localdomain` → `window`):

```js
const sepIdx = bucketId.indexOf('_');
const head = sepIdx === -1 ? bucketId : bucketId.slice(0, sepIdx);
```

When the base id has no underscore, that cut lands inside the origin hostname.

The synced-bucket branch can't catch it, because it requires an underscore *before*
`-synced-from-`:

```js
const syncMatch = bucketId.match(/^([^_]+)_.*-synced-from-(.+)$/);
```

Desktop watcher ids satisfy that by accident
(`aw-watcher-window_HOST-synced-from-ORIGIN`). Ids with no underscore in the base don't, every `aw-watcher-android-*` bucket, and also
`aw-stopwatch-synced-from-my_desktop`, which truncates the same way on desktop.

Android hits it most often because hostnames come from the OS device name with
non-`[a-z0-9_-]` runs replaced by `_`, so "Jude's S25 Ultra" becomes
`jude_s_s25_ultra`.

### Version

master @ `2809ee2`. Regex unchanged since #757 (`1393ec6d`). Related:
ActivityWatch/aw-server-rust#649, which proposes reading provenance from bucket
metadata instead of parsing the id.

TimeToBuildBob:
## Problem

Bucket IDs like `aw-watcher-android-synced-from-my_phone` were rendered incorrectly. The formatted label showed `android-synced-from-my` instead of something like `aw-watcher-android (synced from my_phone)`.

Closes #967

## Root cause

The synced-bucket detection used a regex `/^([^_]+)_.*-synced-from-(.+)$/`:

- `[^_]+` matches characters up to the **first underscore**
- For `aw-watcher-android-synced-from-my_phone`, the first `_` is inside the origin hostname (`my_phone`), so `[^_]+` consumed `aw-watcher-android-synced-from-my`
- The rest was just `phone`, which doesn't contain `-synced-from-`, so the regex never matched
- The code fell through to `shortenBucketLabel()`, which cut at the first `_` to yield `android-synced-from-my`

## Fix

Replace the regex with a direct `indexOf('-synced-from-')` search. This detects the marker regardless of whether the origin hostname contains underscores. The base part's optional `_<hostname>` suffix is then stripped separately (needed for desktop buckets like `aw-watcher-window_host-synced-from-...`).

## Changes

- `src/util/timelineLabels.ts`: replace regex with `indexOf` approach
- `test/unit/timelineLabels.test.node.ts`: add regression for the reported case

All 10 existing tests still pass; the new test covers the previously broken case.

### DIFF
--- a/src/util/timelineLabels.ts
+++ b/src/util/timelineLabels.ts
@@ -49,11 +49,25 @@ export function formatTimelineBucketLabelHtml(
   options: BucketLabelOptions = {}
 ): string {
   const escaped = escapeHtml(bucketId);
-  const syncMatch = bucketId.match(/^([^_]+)_.*-synced-from-(.+)$/);
 
-  if (syncMatch) {
-    const baseLabel = addWrapOpportunities(escapeHtml(syncMatch[1]));
-    const remoteLabel = addWrapOpportunities(escapeHtml(syncMatch[2]));
+  // Detect synced buckets by searching for the literal '-synced-from-' marker
+  // rather than using an underscore-split regex. The old regex
+  //   /^([^_]+)_.*-synced-from-(.+)$/
+  // failed for bucket ids where the origin hostname contains underscores
+  // (e.g. 'aw-watcher-android-synced-from-my_phone'), because [^_]+ stopped
+  // at the underscore inside the hostname, leaving no room for '-synced-from-'
+  // to match in the remaining string.
+  const syncMarker = '-synced-from-';
+  const syncIdx = bucketId.indexOf(syncMarker);
+  if (syncIdx !== -1) {
+    const basePart = bucketId.slice(0, syncIdx);
+    const remotePart = bucketId.slice(syncIdx + syncMarker.length);
+    // Strip optional '_<hostname>' suffix from the base part, which is present
+    // when the source bucket follows the conventional aw-watcher-type_host format.
+    const underscoreIdx = basePart.indexOf('_');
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:

PR #983, opened 2026-09-16
fix(buckets): render device.device_id instead of missing device.id

### LINKED ISSUE #982: Buckets.vue renders `ID: undefined` once data.device_id is populated; device-ID check is dead code
The Raw Data page groups buckets by device and has UI for showing a device ID distinct from the hostname. That UI is unreachable today, and it has a bug that will surface the moment it becomes reachable.

## `device.id` is not a field

`src/views/Buckets.vue`:

```pug
div.small.text-muted(v-if="device.hostname !== device.device_id")
  | ID: {{ device.id }}
```

`bucketsByDevice` in `src/stores/buckets.ts` returns:

```js
return {
  buckets: d,
  device_id: device_ids[0],
  device_ids,
  hostname: hostnames[0],
  hostnames,
  first_seen: ...,
  last_updated: ...,
};
```

There is no `id` key — the field is `device_id`. So the line renders `ID: undefined`.

## Why nobody has seen it

The same getter derives the ID with a hostname fallback:

```js
const device_ids = _.uniq(_.map(d, b => b.data.device_id || b.hostname));
```

and `data.device_id` is never populated in practice. The only code that writes it is the `!local` sentinel branch in `aw-server-rust/aw-server/src/endpoints/bucket.rs`, and no watcher in the bundle uses `!local` — grepping the whole bundle finds only the two server-side implementations. On my own instance:

```
0/32 buckets have data.device_id
```

So `device_id` always equals `hostname`, the `v-if` is never true, and the broken line never renders.

The same fallback disables a consistency check:

```js
{
  msg: () => `Device known by several IDs: ${device.device_ids}`,
  failed: () => device.device_ids.length > 1,
},
```

`device_ids` collapses to a single hostname for every device, so this check cannot fire. Its sibling check on `hostnames` works, because `hostname` is real data.

## Why it matters now

ActivityWatch/activitywatch#302 is about making `device_id` the real identity, and the sync work in ActivityWatch/aw-server-rust#682/#683 ran into exactly the problems that causes — a device whose hostname changed forked into two identities that nothing could reconcile, because the stable ID was nowhere in the data. When `data.device_id` does start being populated, this page is the first place it becomes visible, and today it would show `ID: undefined` alongside a device-ID mismatch warning that has never been exercised.

Cheap to fix while it is still harmless: change `device.id` to `device.device_id`, and give the "several IDs" check a test or at least a manual once-over against a bucket set that actually carries `data.device_id`.

Found while investigating ActivityWatch/aw-server-rust#682/#683/#684.

cc @TimeToBuildBob

TimeToBuildBob:
`bucketsByDevice` returns `device_id`; the Raw Data card interpolated `device.id`. Dead while `data.device_id` is unpopulated (hostname fallback makes the `v-if` never true), and it would show `ID: undefined` once ActivityWatch/activitywatch#302 starts writing real device IDs.

- Template: `device.id` → `device.device_id`
- Test: mismatch branch with a fixture that carries `data.device_id`, plus the several-IDs collection

Does not populate `data.device_id` itself — that's the identity work. Independent of the aw-sync v0.14.0 slice.

Fixes #982


TimeToBuildBob:
## Review guide (auto-posted)

Structured warm-up for reviewers — **what changed**, **what to run**, **where to look**.

### Key files
- `test/unit/store/buckets.test.node.ts` (+76/-0, Δ76)
- `test/unit/Buckets.test.js` (+11/-0, Δ11)
- `src/views/Buckets.vue` (+1/-1, Δ2)

### Suggested verification
- `Run the repo's usual CI-equivalent checks locally`

### Known risks / watch points
- No automatic high-risk tags; use file list + diff for judgment.

### Suggested review focus
- Confirm behavior matches the PR description acceptance criteria.
- Skim the largest diffs first (listed above).
- If CI is green, spot-check the highest-risk paths called out here.

_Generated by `scripts/github/pr-warmup-review-guide.py` for ActivityWatch/aw-webui#983._

TimeToBuildBob:

ErikBjare:
@TimeToBuildBob But https://github.com/ActivityWatch/aw-webui/pull/983#issuecomment-5695179148

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — no P0/P1 findings on latest review

<sub>Updated after inline dispositions on finding threads — this is the current state; the verdict below is frozen at review time and is kept as the historical record of that pass.</sub>

<details><summary>Finding disposition</summary>

| Finding | Severity | State |
| --- | --- | --- |
| `src/stores/buckets.ts:170` | P2 | accepted-tradeoff |
| `src/views/Buckets.vue:21` | P2 | superseded by latest review (not reproduced) |
| `test/unit/Buckets.test.js:8` | P2 | superseded by latest review (not reproduced) |
| `test/unit/store/buckets.test.node.ts:73` | P2 | superseded by latest review (not reproduced) |
| `test/unit/store/buckets.test.node.ts:69` | P2 | accepted-tradeoff |

</details>




The PR changes the Buckets view to display device.device_id instead of the nonexistent device.id, and updates the bucketsByDevice getter to prefer a UUID-shaped device_id over a hostname fallback when grouping buckets by device. It adds two test files: a static source check for the template and a unit test suite for the bucketsByDevice getter.

### Needs a look — P2 only

Confidence 4/5

<details><summary><b>2 findings</b> · ⚠️ 2 P2</summary>

**⚠️ P2 medium** — `src/stores/buckets.ts:176`

The new device_id selection logic in bucketsByDevice can pick a device_id that is not actually a device ID but a hostname from a different group, because the second fallback `device_ids.find(id => !hostnameSet.has(id.toLowerCase()))` only excludes hostnames from the current group. If a bucket in the group has `data.device_id` set to a hostname-like string that is not in the current group's hostnames (e.g., a legacy bucket with `data.device_id: 'other-host'`), and no UUID-shaped ID exists, the getter will return that foreign hostname as the device_id. The Buckets view then shows `ID: other-host` even though that is not a device ID, and the `hostname !== device.device_id` guard makes the label appear, misleading the user. The test 'UUID-shaped device_id wins even over a non-group hostname-like id' only covers the case where a UUID exists, so the fallback path is untested. This is a real correctness issue in the fallback heuristic, not just a missing test, because the code as written will mislabel devices when a non-UUID, non-group-hostname string is present.

```suggestion
Consider also excluding device_ids that match any known hostname from the full bucket list, or only accept UUID-shaped IDs and fall back to the first entry otherwise.
```

<sub>**How this was verified:** Traced the logic: device_ids is built from b.data.device_id || b.hostname. If a bucket has data.device_id='other-host' and no UUID exists, the first find fails, the second find matches 'other-host' because it's not in hostnameSet (which only contains hostnames from the group), so device_id becomes 'other-host'. The view then shows ID: other-host.</sub>

**⚠️ P2 medium** — `test/unit/store/buckets.test.node.ts:69`

The test 'prefers real UUID over hostname fallback in mixed group' uses `data: { device_id: 'real-uuid-xyz' }` which is not a UUID-shaped string (it lacks the 8-4-4-4-12 hyphen pattern). The UUID_RE regex requires a specific format, so 'real-uuid-xyz' does not match. The test expects device_id to be 'real-uuid-xyz', but the code's first find (UUID_RE.test) will fail, then the second find checks if it's in hostnameSet; 'real-uuid-xyz' is not a hostname, so it will be selected by the second fallback. The test passes, but it does not actually test the UUID preference; it tests the non-hostname fallback. The test name and comment claim it verifies UUID preference, but the fixture does not contain a real UUID. This is a test defect: the assertion passes for the wrong reason, so the UUID-preference branch is not actually exercised.

```suggestion
Change the fixture to a real UUID-shaped string, e.g., 'f81d4fae-7dec-11d0-a765-00a0c91e6bf6', to actually exercise the UUID_RE branch.
```

<sub>**How this was verified:** Checked the UUID_RE regex: /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i. 'real-uuid-xyz' does not match because it has letters beyond hex and no hyphens in the right places. The test still passes because the second fallback selects it.</sub>

</details>

<details><summary><b>Files changed (4)</b> — the diff as I read it</summary>

- `src/stores/buckets.ts` — Adds logic in bucketsByDevice to select a device_id preferring UUID-shaped values, then non-hostname values, then the first entry.
- `src/views/Buckets.vue` — Changes the ID label to render device.device_id and adds a guard for device.device_id existence.
- `test/unit/Buckets.test.js` — Adds a static test asserting the template uses device.device_id and not device.id.
- `test/unit/store/buckets.test.node.ts` — Adds unit tests for bucketsByDevice covering hostname fallback, UUID preference, case-insensitivity, and multiple device_ids.

</details>

<details><summary>Previous review passes</summary>

| commit | score | findings | engine | when |
| --- | --- | --- | --- | --- |
| `c9977b376dbc` | 4/5 | 3 | llm | 2026-09-16 09:46 UTC |
| `d2ec3d7c69be` | 4/5 | 2 | llm | 2026-09-16 10:41 UTC |

</details>

<sub>Reviewed `716a184e9ea6` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 23s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

test/unit/Buckets.test.js
```diff
@@ -0,0 +1,11 @@
+const fs = require('fs');
+const path = require('path');
+
+describe('Buckets.vue device ID label', () => {
+  const src = fs.readFileSync(path.join(__dirname, '../../src/views/Buckets.vue'), 'utf8');
+
+  test('renders device.device_id, not the missing device.id field', () => {
+    expect(src).toMatch(/ID: \{\{ device\.device_id \}\}/);
```
TimeToBuildBob, line 8: ⚠️ **P2** — The new test file test/unit/Buckets.test.js is a static source-text test that reads the .vue file and asserts the template contains 'ID: {{ device.device_id }}' and does not contain 'ID: {{ device.id }}'. This test is vacuous as a behavioral test: it only checks the literal source text, not the rendered output. It will pass even if the template is broken in a way that renders nothing or if the v-if condition is wrong, and it will fail if the template is reformatted (e.g., whitespace or line breaks) without any behavioral change. The test does not mount the component or exercise the rendering path, so it provides no signal about whether the device ID label actually appears correctly. This is a test-defect because the assertion is about the source string rather than the component's behavior.

test/unit/store/buckets.test.node.ts
```diff
@@ -0,0 +1,76 @@
...
+      }),
+      bucket({
+        id: 'aw-watcher-afk_erb-m2',
+        type: 'afkstatus',
+        data: { device_id: 'uuid-b' },
+      }),
+    ]);
+
+    const device = Object.values(store.bucketsByDevice)[0];
+    expect(device.device_ids).toEqual(expect.arrayContaining(['uuid-a', 'uuid-b']));
```
TimeToBuildBob, line 133: ⚠️ **P2** — The store test 'collects several device_ids on the same host' asserts that device.device_ids has length 2 and contains both 'uuid-a' and 'uuid-b'. However, the test does not verify that the two buckets are actually grouped into the same device object. The test creates two buckets with the same hostname 'erb-m2' but different types ('currentwindow' and 'afkstatus') and different data.device_id values. If bucketsByDevice groups by hostname, both buckets should be in one device; if it groups by device_id or by bucket type, the test would fail. The test passes only if the grouping is by hostname, which is the intended behavior, but the test does not assert the grouping key explicitly. It also does not check that the device.hostname is 'erb-m2' in this test, so a grouping by device_id would still produce a device with hostname 'erb-m2' (since both buckets have that hostname) but device_ids would be ['uuid-a'] and ['uuid-b'] in separate devices, and Object.values would have two entries, so [0] would only have one device_id, making the length assertion fail. So the test does catch a wrong grouping, but it is not explicit about the grouping contract.

src/views/Buckets.vue
```diff
@@ -19,7 +19,7 @@ div
           span.font-weight-bold {{ device.hostname }}
           b-badge.ml-2(v-if="serverStore.info && serverStore.info.hostname == device.hostname" variant="info") {{ $t('buckets.thisDevice') }}
           div.small.text-muted(v-if="device.hostname !== device.device_id")
```
TimeToBuildBob: ⚠️ **P2** — The template change from device.id to device.device_id is correct for the label, but the v-if condition 'device.hostname !== device.device_id' may be false when device_id is undefined or null. In the fallback case where data.device_id is missing, bucketsByDevice sets device_id to hostname, so the condition is false and the ID line is hidden. That is intended. However, if device_id is undefined (e.g., a device with no hostname and no device_id), the condition 'device.hostname !== undefined' is true, and the template renders 'ID: undefined'. The PR description says this is dead while data.device_id is unpopulated, but the fallback in the store may not cover all cases. The test 'falls back to hostname when data.device_id is missing' covers the case where hostname is set, but not the case where hostname is 'unknown' and device_id is undefined. This could show 'ID: undefined' for unknown devices. This is a bug in the template logic.

TimeToBuildBob:
Fixed in d2ec3d7.

The root cause: when a hostname group had both buckets without `data.device_id` (falling back to hostname) and buckets with a real UUID, `device_ids[0]` could be the hostname fallback. The `hostname !== device_id` guard then hid the label even though a UUID was present in the group.

Fix: prefer any non-hostname entry in `device_ids` before falling back to `device_ids[0]`:
```ts
const device_id = device_ids.find(id => id !== hostname) || device_ids[0];
```

Added a test for the mixed-bucket case (`'prefers real UUID over hostname fallback in mixed group'`).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@ErikBjare You're right — I declared merge-ready while Greptile still had a P1. Mixed groups could pick the hostname fallback as `device.device_id` and hide a UUID that was already in the group. That's d2ec3d7. Greptile re-reviewed 5/5, CI green. Now actually waiting on a merge click.

src/views/Buckets.vue
```diff
@@ -19,7 +19,7 @@ div
...
-            | ID: {{ device.id }}
+            | ID: {{ device.device_id }}
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in d2ec3d7 — mixed groups now prefer a real UUID over the hostname fallback.

src/stores/buckets.ts
```diff
@@ -162,9 +162,15 @@ export const useBucketsStore = defineStore('buckets', {
         d => {
           const hostnames = _.uniq(_.map(d, b => b.hostname || b.data.hostname));
           const device_ids = _.uniq(_.map(d, b => b.data.device_id || b.hostname));
+          // Prefer a real UUID over a hostname fallback: if any device_id in the
+          // group differs from the hostname, use that one. A mixed group (some
+          // buckets without data.device_id) would otherwise surface the hostname
+          // fallback first and hide the ID label in the Buckets view.
+          const hostname = hostnames[0];
+          const device_id = device_ids.find(id => id !== hostname) || device_ids[0];
```
TimeToBuildBob: ⚠️ **P2** — The new device_id selection logic in bucketsByDevice can pick a hostname fallback as the device_id even when a real UUID exists in the group, because the comparison `id !== hostname` is case-sensitive and the hostname fallback is derived from `b.hostname` while the group's `hostname` is derived from `b.hostname || b.data.hostname`. If a bucket's `data.hostname` differs in case from its `hostname` (e.g., hostname 'MyHost' and data.hostname 'myhost'), the fallback device_id 'MyHost' would not equal the group hostname 'myhost', so `find` would return the fallback first and the real UUID would be ignored. This would cause the Buckets view to show the hostname as the device_id, hiding the UUID label. The condition should compare against the actual hostname used for the fallback, or better, filter out values that are equal to any hostname in the group.

test/unit/store/buckets.test.node.ts
```diff
@@ -0,0 +1,136 @@
...
+    // and another has a real UUID. The group device_id must be the UUID so
+    // the Buckets view label is visible (hostname !== device_id guard).
+    const store = useBucketsStore();
+    store.update_buckets([
+      // Older bucket without device_id — would produce hostname fallback
+      bucket({ id: 'aw-watcher-afk_erb-m2', type: 'afkstatus' }),
+      // Newer bucket with a real UUID
+      bucket({
+        id: 'aw-watcher-window_erb-m2',
+        data: { device_id: 'real-uuid-xyz' },
```
TimeToBuildBob, line 69: ⚠️ **P2** — The test 'prefers real UUID over hostname fallback in mixed group' uses `data: { device_id: 'real-uuid-xyz' }` which is not a UUID-shaped string (it lacks the 8-4-4-4-12 hyphen pattern). The UUID_RE regex requires a specific format, so 'real-uuid-xyz' does not match. The test expects device_id to be 'real-uuid-xyz', but the code's first find (UUID_RE.test) will fail, then the second find checks if it's in hostnameSet; 'real-uuid-xyz' is not a hostname, so it will be selected by the second fallback. The test passes, but it does not actually test the UUID preference; it tests the non-hostname fallback. The test name and comment claim it verifies UUID preference, but the fixture does not contain a real UUID. This is a test defect: the assertion passes for the wrong reason, so the UUID-preference branch is not actually exercised.

src/stores/buckets.ts
TimeToBuildBob: Disposition: accepted trade-off — the case-sensitivity half of this finding is already fixed, and the residual behaviour is intended.

- The case-mismatch mechanism was addressed in 1d2562c: the comparison now lowercases every hostname into `hostnameSet` before the fallback `find`, and the thread is outdated against that revision.
- Residual: when the group has no UUID-shaped id, the second fallback can return a non-group, hostname-like `data.device_id` (e.g. `other-host`) and the view renders `ID: other-host`. That is the server-declared `device_id` string; `data.device_id` is an opaque identifier, not guaranteed to be a UUID. Rendering it is the intended behaviour — dropping to `device_ids[0]` instead (the reviewer's suggestion) would hide the label in exactly the mixed-group case this PR exists to fix.
- The UUID-preference branch is covered by `UUID-shaped device_id wins even over a non-group hostname-like id` (uses a real UUID fixture).

No code change: the heuristic as written is the behaviour we want. Converged — remaining findings are P2 only, CI green, mergeable.

test/unit/store/buckets.test.node.ts
TimeToBuildBob: Disposition: accepted trade-off — the test defect is real but does not leave the behaviour unguarded.

- Confirmed: `UUID_RE.test('real-uuid-xyz')` is false, so this test exercises the *second* fallback (`!hostnameSet.has(...)`), not the UUID-preference branch. The name over-claims the path.
- The UUID-preference branch is separately covered by `UUID-shaped device_id wins even over a non-group hostname-like id`, which uses a real UUID fixture (`f81d4fae-7dec-11d0-a765-00a0c91e6bf6`).
- The mixed-group behaviour this test exists for is still genuinely guarded: if the selection regressed to `device_ids[0]`, the assertion `device_id === 'real-uuid-xyz'` would receive `erb-m2` and fail. So the test is imprecise about *which* branch it takes, not vacuous.

Not patching the fixture in this PR — the PR is converged (CI green, mergeable) and a fixture-only commit would restart the review cycle for a naming precision issue. The coverage gap the finding describes does not exist.

TimeToBuildBob:
**Review cycle converged.** Both remaining AI-review threads are now resolved (dispositions posted on-thread, no code change):

- `src/stores/buckets.ts` fallback heuristic — the case-sensitivity half was fixed in `1d2562c`; the residual (rendering a non-UUID `data.device_id` when the group has no UUID) is the intended behaviour — accepted trade-off, not a bug.
- Mixed-group test fixture precision — the UUID-preference branch is covered by the adjacent real-UUID test, so no coverage gap remains.

Status: CI green, `MERGEABLE`/`CLEAN`, 0 unresolved threads. **Maintainer merge only** — Bob has pull-only access on `ActivityWatch/aw-webui`, so this still needs a maintainer click.

TimeToBuildBob:
Merged as `efb8d4dd` — thanks. I rechecked the final state: all 8 checks passed, the mixed-group P1 is fixed, and the linked issue ActivityWatch/aw-webui#982 is closed. Nothing remains on this PR; population of `data.device_id` is the separate ActivityWatch/activitywatch#302 lane.

PR #992, opened 2026-09-20
fix: recover category builder loading on Android

0xbrayo:
The category builder can remain on “Loading…” when bucket discovery, category loading, or the event query fails. Android installations whose only hostname is `unknown` also require an unnecessary manual hostname selection.

Handle initialization and queries through the same error/retry path, select a sole `unknown` host, and ignore superseded responses. Query existing bucket IDs, show an empty state for unsupported hosts, and preserve the selected hostname and dates when reopening options. Desktop AFK filtering remains enabled by default, and ScreenTime retains its title handling.

Validation:
- 363 unit tests pass, including 13 new category-builder regression cases.
- ESLint passes for both changed files.
- Android-targeted webpack build (`make build ON_ANDROID=1`) and Vite build pass.
- Regression tests reproduce failures on the base revision before the fix.

Addresses ActivityWatch/aw-android#123. The Android app will need to consume the merged web UI through its aw-server-rust dependency; this PR does not update those submodule pointers. Android device testing has not been performed for this change.


ErikBjare:
I think the core Category Helper issue has been fixed, but this could be good on top. @0xbrayo ready?

0xbrayo:
@ErikBjare ready for merge, tested :)

